PR #2, opened 2016-05-11
Add support for Python 2.7 and PyPy

Lukasa:
This collection of changes here provides the following:

- Support for Python 2.7
- Support for PyPy
- Support for using tox to run tests.
- Adding Python 2.7 and PyPy to Travis test matrix.


Lukasa:
Looks like codecov isn't counting the Python 2 test runs. Not sure how to handle that.

Lukasa:
Actually, I have a theory. Let's see if that commit helps.

Lukasa:
Yup, got it.

h11/_readers.py
```diff
@@ -236,7 +236,7 @@ def __call__(self, buf):
                 return None
             matches = validate(chunk_header_re, chunk_header)
             # XX FIXME: we discard chunk extensions. Does anyone care?
-            self._bytes_in_chunk = int(matches["chunk_size"], base=16)
+            self._bytes_in_chunk = int(bytes(matches["chunk_size"]), base=16)
```
njsmith: I had to stare at this for a bit (I think it must be because our ReceiveBuffer actually returns bytearrays rather than bytes, and regex matching on a bytearray actually returns bytearray groups, which is surprising but whatever?). So, a comment would be good :-)

test-requirements.txt
```diff
@@ -0,0 +1,3 @@
+pytest
+pytest-cov
+codecov
```
njsmith: I don't have a good handle on conventions around `test-requirements.txt` -- is it normal to have stuff in here that's only needed for CI (codecov), not for running the tests locally?

njsmith:
Sweet! Two tiny comments above, but this is awesome :-)

I guess README and docs/source/index.rst also need updating

njsmith:
I guess I'm also curious now how much work (if any) would be needed to get 2.6 support... though probably not worth caring unless someone actually wants to use it in a project that cares about 2.6

Lukasa:
Ok, ACK to all that feedback.

Regarding Python 2.6, you should very much *not* add support for it. It's painful to do, and while it's definitely *possible*, Python 2.6 is no longer supported by python-dev. That means that if someone wants Python 2.6 support they're going to be using something like RHEL, and you should get either the person asking or Red Hat to *pay* you to add support for Python 2.6.

Don't do all that unnecessary extra work for free. =)

njsmith:
Awesome, thanks!

And yeah, not really planning to spent energy on py26 support -- you notice I didn't even spend energy on py27 support ;-).

(Is RH really paying for requests py26 support?)

Lukasa:
(I wish RH were paying Requests for 2.6 support. Sadly, I consider requests to be the exception here, because pip uses it and pip really does need to keep working. We're not happy about it though.)

njsmith:
Added you as a collaborator too.

Ah, yeah, pip is a problem :-(. RH isn't paying pip for 2.6 support either...

...but you just gave me a great idea for how to get more funding for python packaging infrastructure!

PR #4, opened 2016-05-18
Split receive_data into receive_data + next_event

njsmith:
So now receive_data no longer returns a list of events; instead, it just
appends to the internal buffer, and then you have to call next_event to
get it. (So next_event is like the old receive_data(None).)

Maybe some other small changes snuck in, but that's the main one... this
involved a major rework of the code and docs.

The big advantage of this is that when trying to solve the really hard
problem of implementing a HTTP server that handles errors gracefully,
life is *much* simpler if you can take advantage of h11's knowledge of
the connection state. But with the old way of doing receive_data, you
might get back multiple events at once, and then while you're processing
the first event, conn.states is already updated to reflect the future
events you haven't processed yet. So you couldn't really use the state,
because in general it'd be out-of-sync with your processing.

In the new model, we return events on at a time, and we only update the
state machine as we return them, so now the API follows causal law
again. Down with spooky time travel!

I'm also *much* happier with the flow control section of the docs, which
now explicitly talks about how to use these primitives to implement both
push-style and pull-style servers.

Plus: super-awesome new example curio-server.py, which easily scales to
thousands of requests/second, thousands of concurrent connections, and
has robust error handling without spaghetti (!!).

### DIFF
--- a/docs/source/_examples/myclient.py
+++ b/docs/source/_examples/myclient.py
@@ -18,6 +18,14 @@ class MyHttpClient:
             else:
                 self.sock.sendall(data)
 
-    # max_bytes set intentionally small for pedagogical purposes
-    def receive(self, max_bytes=200):
-        return self.conn.receive_data(self.sock.recv(max_bytes))
+    # max_bytes_per_recv intentionally set low for pedagogical purposes
+    def next_event(self, max_bytes_per_recv=200):
+        while True:
+            # If we already have a complete event buffered internally, just
+            # return that. Otherwise, read some data, add it to the internal
+            # buffer, and then try again.
+            event = self.conn.next_event()
+            if event is h11.NEED_DATA:
+                self.conn.receive_data(self.sock.recv(max_bytes_per_recv))
+                continue
+            return event
--- a/docs/source/api.rst
+++ b/docs/source/api.rst
@@ -16,17 +16,18 @@ directly at the top level:
 
    @verbatim
    In [3]: h11.<TAB>
-   h11.CLIENT                 h11.MIGHT_SWITCH_PROTOCOL
-   h11.CLOSED                 h11.MUST_CLOSE
-   h11.Connection             h11.PRODUCT_ID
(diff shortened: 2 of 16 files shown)


njsmith:
@Lukasa: Not asking for a code review, but you might be interested in this -- it implements a fundamental change in how the h11 receive API works.

Probably the simplest way to understand what + why is to look at the [core of the new curio-server.py example](https://github.com/njsmith/h11/blob/1d9cd6eb00e3de634e1ff426894cf78f40ac7e31/examples/curio-server.py#L190) and how it's a super-elegant little graceful-error-handling thing... BUT totally dependent on having a consistent view of the connection's internal state, which wasn't possible with the old API that returned multiple received events at a time.

Ironically this means that h11 now has a split API for receiving bytes from receiving events, which is exactly opposite and parallel to how h2 has a split API for sending events from sending bytes... :-/

Lukasa:
> Ironically this means that h11 now has a split API for receiving bytes from receiving events, which is exactly opposite and parallel to how h2 has a split API for sending events from sending bytes... :-/

Heh, yeah, I clocked that. I thought that was pretty funny. ;)

So, this does touch on a constant concern I've had, which is that it's possible that the state machine will reject actions you take on it based on events if the user hasn't yet handled certain stream-state-affecting events (e.g. RST_STREAM, GOAWAY).

I'm honestly not sure how best to handle that concern. In the case of h2, at least, if you try to send data on a stream that got reset, nothing bad will happen aside from the exception that gets thrown. That seems *ok* to me now, but I'm definitely not certain about it.

Interestingly, however, h2 has a perfect solution for dealing with this problem. Right now `receive_data` returns a list, but there's no reason it couldn't return a generator. If it did, we'd be able to make the guarantee that the state machine reflects the state *only* for the frame that was just processed.

I'm not 100% sure that this is a good idea for h2, but it's worth discussing. I'll open an issue on h2 and try to think it over with some h2 contributors as well.

PR #6, opened 2016-05-31
Strip Outgoing header values

### LINKED ISSUE #5: possibly should be stricter about checking outgoing headers
Right now we e.g. allow newlines inside headers, which is kinda bad. (Could even conceivably be a security risk if untrusted code is allowed to set header values.)

accraze:
This PR contains changes to `_writers.py`:
Removes trailing characters and spaces from outgoing header values.

Resolves #5

### DIFF
--- a/h11/_writers.py
+++ b/h11/_writers.py
@@ -31,7 +31,7 @@ else:
 
 def write_headers(headers, write):
     for name, value in headers:
-        write(bytesmod(b"%s: %s\r\n", (name, value)))
+        write(bytesmod(b"%s: %s\r\n", (name, value.strip())))
     write(b"\r\n")
 
 # XX FIXME: "Since the Host field-value is critical information for


njsmith:
Hi @accraze -- thanks for your interest and contribution!

Unfortunately, there are a few issues:

- This doesn't actually address issue #5, which is about doing more careful checking of the contents of outgoing header values. The example given there is a newline *inside* a header value, which is unaffected by stripping whitespace around the edges of a header value. But there are also other ways that a header value can be invalid beyond just containing embedded newlines.

- No test is included -- as noted in [our contribution guidelines](https://github.com/njsmith/h11/blob/master/CONTRIBUTING.md), all functionality changes need some corresponding change to the test suite. (Though now that I look at it that section probably isn't phrased as clearly as it could be... sorry about that.)

Also, I'm actually experimenting with a different general strategy for handling headers and header validation on a local branch right now... which of course you had no way to know :-). But I'd prefer to leave this particular issue alone for now until I've gotten a bit further on that, so I'm going to close this for now. Hopefully the above notes are useful for future reference, though.

Issue #8, opened 2016-07-02
Give sentinel objects (or at least NEED_DATA and PAUSED) unique types


njsmith:
@dabeaz has requested this, to make it easier to take the thing that comes out of `next_event()` and dispatch it using `functools.singledispatch` or something like:

```
 class Handler(object):
     def handle_events(self):
            while True:
                   evt = conn.next_event()
                   name = 'handle_' + type(evt).__name__
                   getattr(self, name)(evt)

     def handle_Data(self, evt):
           ...

     def handle_EndOfMessage(self, evt):
          ...

     def handle_Response(self, evt):
          ...

     def handle_NEED_DATA(self, evt):
          ...
```

Seems reasonable enough to me.

njsmith:
Fixed by #18

PR #11, opened 2016-08-06
Unique types for sentinel objects

RazerM:
Proposed solution to #8.

In this implementation, an empty `Sentinel` class replaces the existing one, which unique sentinel types then inherit from (so that `Connection.next_event` can still check whether an event is an instance of `Sentinel`).

A `make_sentinel` function is added to create the type and return an instance.


njsmith:
See also #18, which is this PR + a somewhat wacky trick added.

njsmith:
#18 is merged, so closing this.

PR #13, opened 2016-08-26
Add reason at Response

chhsiao90:
Though there is useless for "reason" on the Response.
but I think keep it still better than discard when using reader to parse it.
So I create a PR to keep the reason on Response.

I had following modification:
1. Retrieve reason when using reader to parse the incoming data by modify the regex
2. Give reason default value empty string if the server don't care about it and will send empty string to client.

BTW, I know almost every client/server don't care about it,
but for my usage, my friend and I had implement a proxy server that was using h11,
and try keep original message that generated by client/server to prevent any misunderstand.
I had another modification on h11 that is related to it: keep original header field case.
I will create another PR after your feedback if you like it.

Thanks!

### DIFF
--- a/h11/_events.py
+++ b/h11/_events.py
@@ -44,7 +44,7 @@ class _EventBundle(object):
         if "headers" in self.__dict__:
             self.headers = _headers.normalize_and_validate(self.headers)
 
-        for field in ["method", "target", "http_version"]:
+        for field in ["method", "target", "http_version", "reason"]:
             if field in self.__dict__:
                 self.__dict__[field] = bytesify(self.__dict__[field])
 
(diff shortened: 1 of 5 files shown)


njsmith:
Meh, I'm not a fan of this part of HTTP, but I suppose you're right :-). Patch looks fine, but could you also update the docs to match?

Regarding header case: I am not sure that this is a good trade-off... it adds a lot of easy-to-mess-up complexity to user code if they have to worry about case normalization at use time. Note that nodejs's http stack unconditionally lowercases header names and they seem to manage OK.

(Also, as a heads up, I've been considering some much more radical changes to the header handling/representation, since the current code is really cumbersome in lots of situations.)

chhsiao90:
I will update the pydoc for this PR, actually I had planned to do this after your feedback!

For header part, I agreed that the solution is ugly, too. So I will re-design it.

Thanks for your feedback!

chhsiao90:
I had updated the doc for field reason on Response/InformationalResponse.
Please help check that the term I used is properly or is there any I could improved.
Thanks!

njsmith:
Looks good, thanks!

Issue #14, opened 2016-08-26
_EventBundle reports wrong kwarg name when required kwarg is missing.


jeamland:
Python 3.5.2 (default, Aug 16 2016, 05:35:40)
    [GCC 4.2.1 Compatible Apple LLVM 7.3.0 (clang-703.0.31)] on darwin
    Type "help", "copyright", "credits" or "license" for more information.
    >>> import h11
    >>> h11.Response(status_code=200)
    Traceback (most recent call last):
      File "<stdin>", line 1, in <module>
      File "/Users/riceb2/.pyenv/versions/nodinator/lib/python3.5/site-packages/h11/_events.py", line 41, in __init__
        .format(kwarg, self.__class__.__name__))
    TypeError: missing required kwarg status_code for Response
    >>> h11.Response(headers={})
    Traceback (most recent call last):
      File "<stdin>", line 1, in <module>
      File "/Users/riceb2/.pyenv/versions/nodinator/lib/python3.5/site-packages/h11/_events.py", line 41, in __init__
        .format(kwarg, self.__class__.__name__))
    TypeError: missing required kwarg headers for Response
    >>> h11.Response(status_code=200, headers={})
    Response(status_code=200, headers=[], http_version=b'1.1')
    >>>

njsmith:
Fixed by #15

PR #15, opened 2016-08-26
Use correct variable when telling the user which kwarg they forgot.

jeamland:
Fixes issue #14.


njsmith:
Oh good catch, thanks!

Could you also add a test? (Probably something like: find the test that currently checks that a `TypeError` is raised here and add a `assert "kwarg name" in str(exc)`)

jeamland:
I've added the test, let me know if it needs anything else.

Thanks!

njsmith:
Thanks!

Issue #23, opened 2016-12-01
Short requests/responses could use more granular exceptions.


Lukasa:
When either the server or the client sends a HTTP message with a body that is *short* (that is, the body does not contain enough bytes to meet the `Content-Length` header, or the chunked encoding is aborted without the ending), an exception like this is raised:

```
RemoteProtocolError("can't handle event type <class 'h11._events.ConnectionClosed'> for SERVER in state SEND_BODY",)
```

This message, while understandable, is a little hard to introspect. It would be nice if we could have a subclass of the `ProtocolError` that indicated exactly the kind of error, so that applications that would like to report these errors in a more granular way could do so.

njsmith:
Sure, it's hard to object to finer-grained errors :-).

What exactly do you want to detect? is it specifically: any time the state machine errors out due to an unexpected event, where that event is of type `ConnectionClosed`? Do you also want this if they drop the connection when, say, they're half-way through the headers?

njsmith:
(NB I just also just tweaked that error message slightly in master, because `<class h11._events.ConnectionClosed>` was bugging me now that you pointed it out :-))

Lukasa:
@njsmith Something like that, yeah. Basically, I'd like to be able to reliably detect violations in HTTP framing that have to do with the body: whether the body is too short or too long.

Halfway through the headers bothers me less *right now*, though I may come back to it at some point. This is part of my work on urllib3, which has historically thrown specific errors for bodies that are too short/too long.

njsmith:
Oh, I see, right, that's a little different.

How do you detect bodies that are too long? Right now I think h11 will just buffer any "extra" bytes and treat them as part of the next request/response cycle...

For too short, it sounds like you specifically want to detect `ConnectionClosed` when in state `SEND_BODY`. As a wrinkle, though, note that if a chunked-encoded connection gets dropped, we aren't actually guaranteed to get that specific exception you noticed, with the invalid state transition. We only detect that state transition when the parser reaches the EOF, and that only happens if we successfully parse all the data leading up to the EOF. If the connection drops half-way through a chunk header (like the last few bytes are `\r\n100\r` then EOF), then we instead hit [this branch](https://github.com/njsmith/h11/blob/421d32229e114f32acf7b3e80b0b5d1c3bbdfd69/h11/_connection.py#L430-L434), which detects the generic condition "the parser is waiting for more data but the connection is closed".

Some possible approaches:

[...]

PR #25, opened 2016-12-03
More detailed errors when connection closed during message body

njsmith:
Intended as at least a first step towards gh-23.

### DIFF
--- a/h11/_readers.py
+++ b/h11/_readers.py
@@ -17,7 +17,7 @@
 # - or, for body readers, a dict of per-framing reader factories
 
 import re
-from ._util import LocalProtocolError, validate
+from ._util import LocalProtocolError, RemoteProtocolError, validate
 from ._state import *
 from ._events import *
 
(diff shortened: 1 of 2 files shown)


njsmith:
@Lukasa: so here's a first step towards fixing #23, with "human readable" errors only -- I guess we can merge this first and then decide whether to add special exception types as well.

The exception message phrasing currently a bit awkward IMO -- suggestions welcome.

njsmith:
Oh, and to be clear, the full behavior with this PR is:

* connection dropped during content-length body: "peer closed connection without sending complete message body (received N bytes, expected M)"

* connection dropped during a chunked body while we are waiting for the `\r\n` at the end of a chunk header, or while reading trailers: "peer unexpectedly closed connection" (generic fallback)

* connection dropped during chunked body, at any other moment: "peer closed connection without sending complete message body (incompleted chunked read)"

njsmith:
Cool, let's do this then. (As usual, let me know if you need me to roll a release.)

PR #27, opened 2016-12-12
restrict newlines in header values

nateprewitt:
This should address #5 by raising an exception in the event of return characters in the header field value. I also added an exception for requests sending both Content-Length and Transfer-Encoding. [RFC 7230 3.3.2](https://tools.ietf.org/html/rfc7230#section-3.3.2) is pretty clear about that, so I felt like it was a safe call.

The only other major issues I can think of for cleaning up outgoing headers are already covered. Leading/trailing whitespace is taken care of by the `strip` on value, and trailing space on the name is handled the same way.

Let me know if I'm overlooking anything else.

### DIFF
--- a/h11/_headers.py
+++ b/h11/_headers.py
@@ -56,6 +56,7 @@ from ._util import LocalProtocolError, bytesify, validate
 # Maybe a dict-of-lists would be better?
 
 _content_length_re = re.compile(br"[0-9]+")
+_header_value_re = re.compile(br"[\x21-\xff]+(?:[ \t]+[\x21-\xff]+)*|^$")
 
 def normalize_and_validate(headers):
     new_headers = []
@@ -64,6 +65,9 @@ def normalize_and_validate(headers):
     for name, value in headers:
         name = bytesify(name).lower()
         value = bytesify(value).strip()
+        # Ensure header value doesn't contain any non-permitted characters
+        validate(_header_value_re, value,
+                 "Illegal header value {}".format(value))
         # "No whitespace is allowed between the header field-name and colon.
         # In the past, differences in the handling of such whitespace have led
         # to security vulnerabilities in request routing and response
--- a/h11/tests/test_headers.py
+++ b/h11/tests/test_headers.py
@@ -15,6 +15,14 @@ def test_normalize_and_validate():
     # leading/trailing whitespace on values is stripped
     assert normalize_and_validate([("foo", "   bar  ")]) == [(b"foo", b"bar")]
 
+    # no return or NUL characters in values
+    with pytest.raises(LocalProtocolError):
+        normalize_and_validate([("foo", "bar\r\nbaz")])
+    with pytest.raises(LocalProtocolError):
+        normalize_and_validate([("foo", "bar\nbaz")])
+    with pytest.raises(LocalProtocolError):
+        normalize_and_validate([("foo", "bar\x00baz")])
+
     # content-length
     assert (normalize_and_validate([("Content-Length", "1")])
             == [(b"content-length", b"1")])


h11/tests/test_connection.py
```diff
@@ -59,7 +59,7 @@ def req(cl=None, te=False):
             assert _body_framing(meth, r) == ("content-length", (0,))
 
     # Transfer-encoding
-    for kwargs in [{"te": True}, {"cl": 100, "te": True}]:
```
nateprewitt: I removed this from the tests for `_body_framing` because it doesn't appear to be testing anything in particular and was conflicts with the new changes. `_body_framing` returns as soon as it hits 'Transfer-Encoding' so 'Content-Length' was never reached in this test anyway.

njsmith:
Awesome, thanks for working on this!

On a quick skim I noted a few high-level issues, so I'll just mention those:

I think I'm fine with disallowing `Content-Length`+`Transfer-Encoding` on outgoing messages, but if we're doing this then there's some code in `Connection._clean_up_response_headers_for_sending` that also needs changing. (I'm surprised you didn't hit it in the tests, actually... I guess I must have missed writing a test for it? Eek.) We do still need to handle the combination on incoming messages as per the RFC, of course, and to make sure we're still testing that.

The right way to validate the headers is to take the regexes for `field-name` and `field-content` (I think) from `_readers.py`, and check against them. Those are the formal definition of what's allowed in a header name/value, and it's a little bit more subtle than just strip+no-`\r\-or-`\n` :-). (E.g., NUL is also illegal.)

h11/tests/test_connection.py
njsmith: Well, it's testing that `_body_framing` correctly ignored the `Content-Length` when `Transfer-Encoding` was provided :-). We don't necessarily need to test it right here in this way, but it is a useful thing to test.

h11/_headers.py
```diff
@@ -74,6 +74,11 @@ def normalize_and_validate(headers):
         # downstream." -- https://tools.ietf.org/html/rfc7230#section-3.2.4
         if name.strip() != name:
             raise LocalProtocolError("Illegal header name {}".format(name))
+        # "A sender MUST NOT generate a message that includes line folding"
+        # -- https://tools.ietf.org/html/rfc7230#section-3.2.4
```
njsmith: That's not really the relevant section -- it's saying that when encoding headers into the wire format, you shouldn't use line folding as part of the encoding. The rule we're enforcing here is that the header payload (the thing that gets encoded -- think like unicode strings versus byte strings) cannot contain `\r` or `\n` (among other things).

njsmith:
...Ugh, in fact it looks like we might not have any end-to-end tests that feed in a header string with both `Content-Length` and `Transfer-Encoding` in it? Because I think right now `_validate` is called for both incoming and outgoing headers equally, so these changes actually are probably breaking the incoming case, which is worrisome (the tests should have caught this!) and annoying because the fix will probably make things more complicated. I guess we need different validation for incoming and outgoing headers? 

(This could actually be a blessing in disguise since it means that we can optimize by not re-applying regex validation to incoming headers when we have just used those exact same regexes to parse the headers in the first place -- I've debated whether to add such a split in order to get this optimization in the past already.)

Lukasa:
> I guess we need different validation for incoming and outgoing headers?

Almost certainly. ;)

So, h2 already does this, and h11 will need to do it even more. The main problem here is that in the HTTP/2 community right now there is a consensus that strictly enforcing the protocol is the way to go, but in HTTP/1.1 all bets are off. That means h11 really wants two policies: it wants to be *very strict* about the header blocks it emits (basically being as sure as possible that they're spec compliant), but be substantially more relaxed about the header blocks it receives when those header blocks can be understood.

nateprewitt:
Alright, so there's a lot here! You're completely right @njsmith that this will break incoming Requests. I wasn't thinking about this as a bidirectional library but rather only a client when writing the tests. Shows how careful I am at reading docs :)

So actionable items, we're probably going to need to divide validation into incoming and outgoing like we're doing in h2. That I'm comfortable with, but there's a question of conveying directionality in the event. Right now it looks like knowledge of whether the instance is a SERVER or CLIENT is restricted to the connection. Do you have a problem piping that down into the event objects? We use `is_client ` as a header validation flag in h2 presently.

The other outstanding question in my mind is where do we want these validations to live? I feel like `_validate` provides use on a per object basis for specific edge cases. Would two generic `normalize_and_validate_outgoing` vs `normalize_and_validate_incoming` in _headers.py be acceptable for broader stuff?

I'll try to marinate in the code a little bit today so I can get a better feeling on how things are working before I start. Let me know if either of you have particular vision for what this looks like to you.

h11/_headers.py
nateprewitt: Hrmm, while I'm sure the other language is there in the RFC, I am struggling to find it at the moment. I went back and checked the tickets for when we did this in Requests and didn't find a specific reference there either.

Do you or @Lukasa know off hand which section the verbage is in, so we can reference it correctly? Otherwise, I'll take a deeper look tonight.

Lukasa: There is no other section.

The reason we want to forbid newlines is not because they're not RFC-valid: as @njsmith points out, they are. The reason we want to forbid them is because they can be used to conduct request smuggling attacks if users are in any way capable of inserting data into the header fields. This should be forbidden. Given that line folding is forbidden *anyway*, and that h11 does not allow users to control the way HTTP is framed (that's what h11 does!), the whitespace characters amount to framing characters and as such should be forbidden.

nateprewitt: Ok cool, that was my understanding from our first pass in Requests. Just wanted to clarify I wasn't missing a more direct mandate.

njsmith:
This isn't about client versus server, it's about us versus them -- we don't want to send Content-Length+Transfer-Encoding (as either a client or server), but we should allow receiving it (as either a client or server).

Looking again, I should clarify that the way h11 has avoided this so far actually does make some sense: right now we *allow* users to say "please send this header block that has both Content-Length and Transfer-Encoding in it", but if they do then we throw away the Content-Length header before sending it, so our actual wire data is always spec-compliant. This is consistent with how the spec suggests proxies should behave ("A sender MUST remove the received Content-Length field prior to forwarding such a message downstream"), and consistent with our general strategy for munging these headers -- we also generate or discard TE headers in some cases.

(Basically our approach right now is: when the library user gives us some headers to send, we apply the RFC 7230 rules to figure out whether the message body is fixed-length or streaming, as if these headers were something received over the wire from an HTTP/1.1 peer. Then we generate whatever headers are most appropriate for indicating a fixed-length or streaming message when talking to the peer we actually have. This is a Good Thing for the Transfer-Encoding header because HTTP/1.0 peers don't support Transfer-Encoding so you should never use it, while HTTP/1.1 peers do support the HTTP/1.0-style of delimiting the message body with EOF, but you should never use it and always use Transfer-Encoding instead, and asking library users to keep track of this is clearly silly; so instead, users can just use either the HTTP/1.0 or HTTP/1.1 headers and then we fix them up. And currently the Content-Length handling -- which is more debatable -- gets swept up in the same basic logic. So do some other rules like: "A server MUST NOT send a Transfer-Encoding header field in any response with a status code of 1xx (Informational) or 204 (No Content).  A server MUST NOT send a Transfer-Encoding header field in any 2xx (Successful) response to a CONNECT request (Section 4.3.6 of [RFC7231]).)")

So there's a deeper question here about whether h11 should continue with this pattern of accepting the same thing from the library user as we would from a remote peer and applying the same normalizations, or whether we should be require library users to "pre-normalize" when possible. Other potential examples:

* The spec allows receivers to normalize `Content-Length: 42, , 42` into `Content-Length: 42` instead of rejecting it. Right now we reject it, but I can imagine compatibilit pressure might force us to relax this at some point.

* We also bytesify and lowercase user-provided headers names; we could instead require them to be bytes-only and lowercase-only when provided. (I did it this way because by the time h11 has paid the cost to check if they're valid, fixing them is basically free -- but I'm not sure I made the right decision.)

* We bytesify and strip user-provided header values. Stripping header values is similar to lowercasing header names, but maybe even more dubious.

njsmith:
Reviewing this... to move forward, I suggest that you make this PR just the header content validation, and then we should split off the Content-Length-related discussion to a new issue.

nateprewitt:
Sorry I wasn't clear, I was referring to client vs server as context for determining what is "outgoing" (request vs response). Similar to how you use `our_role` vs `their_role`. I agree though this is probably a larger conversation for elsewhere.

I'll move over to using the regex from value validation and rebase the PR to just that :)

Lukasa:
> So there's a deeper question here about whether h11 should continue with this pattern of accepting the same thing from the library user as we would from a remote peer and applying the same normalizations, or whether we should be require library users to "pre-normalize" when possible. 

My recommendation is that you normalize as much as possible on the user's behalf (where there is no ambiguity), and then that you ask that users pre-normalize for anything else.

In the case of the C-L/T-E headers, I think that h11's current behaviour is quite reasonable. Ultimately, I consider those headers to be a *framing* tool, and so I think it's fair game for h11 to manipulate them somewhat rather than simply validating them: after all, ultimately it's h11 that will be held accountable for policing them and framing the data correctly!

However, in general I think h11 should refuse to send headers that are ill formed. For example, h11 should police `Content-Length` for reasonableness, and if it doesn't believe the header is reasonable and it also cannot be mended, then h11 should feel quite happy to refuse to emit it.

nateprewitt:
Alright @njsmith, I think this should be revised appropriately whenever you have a free moment :)

nateprewitt:
Just another data point here for ingoing vs outgoing headers. I'm looking at the project list for urllib3 v2 and it looks like we're trying to remove our regexes there, if h11 covers it. This patch allows h11 to take over for values, but we still have some cases for names we're not checking. I don't know if we want to replace `name.strip() != name` with the `token` regex on incoming headers, but we'll probably want that for all outgoing headers.

njsmith:
Incoming header names and values are already validated for ABNF-compliance, because we use those regexes to parse them. (I.e. if the header name isn't a valid `token` then we'll just fail to parse that header line entirely and error out.) The stripping in `_validate_headers` is only relevant for locally-created headers.

h11/_headers.py
```diff
@@ -56,6 +56,7 @@
 # Maybe a dict-of-lists would be better?
 
 _content_length_re = re.compile(br"[0-9]+")
+_header_value_re = re.compile(br"[\x21-\xff]+(?:[ \t]+[\x21-\xff]+)*|^$")
```
njsmith: Would it make more sense to do `from ._readers import field_value` instead of duplicating the regex here?

nateprewitt: That was my first choice, but we get a bit of an import mess because _readers, _state, _and _event all import * from each other.

_state is dependent upon _connection being imported before it, which is handled in __init__.py. However, _events is imported first in __init__ which causes the problem. _events imports _headers, which will then import from _readers which imports everything from _state. _state fails because it doesn't have what it needs from _connection, which hasn't been imported yet.

If the definitions are going to be nicely reusable, they'll likely need to be moved into either a utils or definitions file.

njsmith:
Thanks for this, and sorry for the delay! I went ahead and refactored a bit more on master after merging, to remove the duplicate regex and add a similar test for header names.

PR #46, opened 2017-04-26
Adapting curio-server.py to trio

merrellb:
Combining my two favorite projects by @njsmith


njsmith:
Now that Trio finally has a solid set of highlevel networking APIs, I just updated this to use them and switched over the examples to using Trio consistently. Now let's see what Travis thinks...

njsmith:
...or maybe this is a doc-only change and we don't currently test docs, plus travis is down anyway. YOLO!

jmehnle:
I wish you hadn't removed the `curio` example. It's useful.

Issue #55, opened 2018-02-04
Allow manual transition to Error state


Nikratio:
I am concerned about the situation where `Connection.send` gives me some data, but I am not able to actually send the data to the peer (e.g. because the connection was closed).

The cleanest solution that does not need any additional code higher up in the stack seemed to be to raise `h11.RemoteProtocolError` and switch the connection state to error. However, the connection state seems to be a read-only attribute.  This is unfortunate, because if I can't switch the connection state then I need to propagate this information differently - which is really weird because that means the same event (peer closed connection) will randomly exercise two different code paths (depending on the event being detected when we attempt to send or receive).

Would it be possible to add an API function to force the connection into error state? Or alternatively, some way to communicate a failed send() back into h11?

njsmith:
Yeah, urllib3 needs this too: https://github.com/njsmith/urllib3/blob/1ca67ee53e18f823d0cbc1b6a0fbbe9f620bebc3/urllib3/_async/connection.py#L233-L242

(Please don't copy that code there, that's a temporary hack.)

It would be good to have some method like `conn.send_failed()` or something though, yeah. (Possibly implemented like the temporary hack.) Interested in putting together a PR?

Nikratio:
I would certainly like to contribute, yeah (same goes for the trio and pytest-trio features that I'm interested in). But unfortunately I don't think I'll be able to. It's the eternal open-source problem: too many potential projects with too few people working on it. I try to spend the little time that I have available in the areas where I can have the most impact, which unfortunately are the projects where I am already familiar with the codebase. I don't have time to become regular contributor to another project, and the (unproductive) ramp-up time required for one-off contributions like this just too high.

I'm sure you are facing the same choices, so I hope you understand.

PR #56, opened 2018-03-20
Add Connection.send_failed

njsmith:
This is useful for urllib3, and fixes gh-55

### DIFF
--- a/docs/source/api.rst
+++ b/docs/source/api.rst
@@ -383,6 +383,7 @@ The Connection object
    .. automethod:: next_event
    .. automethod:: send
    .. automethod:: send_with_data_passthrough
+   .. automethod:: send_failed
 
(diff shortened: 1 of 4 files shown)


njsmith:
CC @pquentin

pquentin:
Should we add a purple line with send_failed to the diagram, like it's done for start_next_cycle? Or it's not useful/too noisy here?

njsmith:
The text below the diagram already says:

"That weird ERROR state hanging out all lonely on the bottom: to avoid cluttering the diagram, we don’t draw any arcs coming into this node, but that doesn’t mean it can’t be entered. In fact, it can be entered from any state: ..."

So I think we're good on that.

Thanks for the review! Going to release 0.8.0 now...

Issue #57, opened 2018-03-20
Fails to decode a header that requests can handle


miracle2k:
I am trying to read this url

`https://www.bitstamp.net/api/v2/trading-pairs-info/`

It fails (see https://github.com/theelous3/asks/issues/60) with an exception:

```
....
  File "/Users/michael/.pyenv/versions/3.6.3/lib/python3.6/site-packages/h11/_readers.py", line 85, in maybe_read_from_SEND_RESPONSE_server
    return class_(headers=list(_decode_header_lines(lines[1:])),
  File "/Users/michael/.pyenv/versions/3.6.3/lib/python3.6/site-packages/h11/_readers.py", line 55, in _decode_header_lines
    matches = validate(header_field_re, line)
  File "/Users/michael/.pyenv/versions/3.6.3/lib/python3.6/site-packages/h11/_util.py", line 96, in validate
    raise LocalProtocolError(msg)
```

Looking closer, this is the header that seems to cause trouble:

```bytearray(b'Set-Cookie: ___utmvafIumyLc=kUd\x01UpAt; path=/; Max-Age=900')```

My guess is it's the ``\x``.

njsmith:
Yep, RFC 7230 says unambiguously that you can't have characters below `\x20` in header values (except for `\t`), so that site is violating the HTTP spec.

Now the question is... what to do about it. If they're violating the spec like that and getting away with it, then probably others are as well, so we might need to relax h11 to match. I'm not keen on allowing just any character inside headers though; that seems like it will inevitably create security issues. I guess we have to go figure out what characters browsers and curl allow...

njsmith:
Things I've learned so far:

* The HTTP/2 RFC [says](https://tools.ietf.org/html/rfc7540#section-10.3): "Any request or response that contains a character not permitted in a header field value MUST be treated as malformed", because they're worried about security attacks where different steps in the chain process these things differently. They particularly call out CR, LF, and NUL as likely to be dangerous. Unfortunately the HTTP/1.1 RFC doesn't get as specific about this point.

* Golang: has a `ValidHeaderFieldValue` check that seems to correctly implement RFC 7230's rules, and if we [look at where it's called](https://golang.org/search?q=ValidHeaderFieldValue), then it looks like they enforce it in their HTTP/1 server code (i.e., reject requests that contain invalid headers), but not in their client code. Also their HTTP/2 code checks this at several places, but I'm not sure how those map onto request/response paths.

[...]

Issue #58, opened 2018-03-20
Bug in vchar_or_obs_text regex: shouldn't accept DEL


njsmith:
Currently our `vchar_or_obs_text` regex is:

```python
vchar_or_obs_text = r"[\x21-\xff]"
```

However, this is incorrect... VCHAR is `%x21-7E` and obs-text is `%x80-FF`, so there's actually a one character gap in the middle of the range: `\x7f` (DEL) should *not* be included.

Of course, we might need to relax these checks anyway (#57), but if we do this should be intentional, not an accident :-)

njsmith:
We decided to redefine `vchar_or_obs_text` to allow most control characters (see #57 and #68), so this has been declared Not A Bug.

PR #68, opened 2018-05-06
Allow more characters in header values

njsmith:
The RFC says we should reject any header value that contains control
characters. But apparently in the real world, you have to both accept
and produce these sometimes (e.g. Google Analytics cookies use them).

As a compromise, we now accept most control characters, but continue
to disallow NUL (\x00) and all whitespace (\t\n\r\f\v and space),
except that space and tab are allowed inside header values when
surrounded by non-whitespace characters.

Closes: gh-57, gh-58


njsmith:
@sigmavirus24 thanks!

sigmavirus24:
You're welcome

PR #69, opened 2018-07-30
Constrain regex request-target matching

pgjones:
This constrains the matching to the printable ASCII characters that
are not the space ' ' character i.e. between '!' (65, 0x21) and '~'
(126, 0x7e). Doing so will mean that a request containing bytes
outside of this range will not match and instead raise an error.

This is based on my understanding of RFC 3986.

I think I've matched the style of the file, although I'm happy for this to be edited however to fit.

### DIFF
--- a/h11/_events.py
+++ b/h11/_events.py
@@ -5,8 +5,10 @@
 #
 # Don't subclass these. Stuff will break.
 
+import re
 from . import _headers
-from ._util import bytesify, LocalProtocolError
+from ._abnf import request_target
+from ._util import bytesify, LocalProtocolError, validate
 
 # Everything in __all__ gets re-exported as part of the h11 public API.
 __all__ = [
@@ -18,6 +20,7 @@ __all__ = [
     "ConnectionClosed",
 ]
 
+request_target_re = re.compile(request_target.encode("ascii"))
 
 class _EventBundle(object):
     _fields = []
@@ -130,6 +133,8 @@ class Request(_EventBundle):
             raise LocalProtocolError("Missing mandatory Host: header")
         if host_count > 1:
             raise LocalProtocolError("Found multiple Host: headers")
+
+        validate(request_target_re, self.target, "Illegal target characters")
 
 
 class _ResponseBase(_EventBundle):
--- a/h11/tests/test_events.py
+++ b/h11/tests/test_events.py
@@ -94,6 +94,15 @@ def test_events():
             headers=[("Host", "a"), ("Foo", "asd\x01\x02\x7f")],
             http_version="1.0")
 
+    # Request target is validated
+    for bad_char in b"\x00\x20\x7f\xee":
+        target = bytearray(b"/")
+        target.append(bad_char)
+        with pytest.raises(LocalProtocolError):
+            Request(method="GET", target=target,
+                    headers=[("Host", "a")],
+                    http_version="1.1")
+
     ir = InformationalResponse(status_code=100, headers=[("Host", "a")])
     assert ir.status_code == 100
     assert ir.headers == [(b"host", b"a")]
--- a/h11/tests/test_io.py
+++ b/h11/tests/test_io.py
@@ -388,6 +388,14 @@ def test_reject_garbage_in_header_line():
            b"Host: foo\x00bar\r\n\r\n",
            None)
 
+def test_reject_non_vchar_in_path():
+    for bad_char in b"\x00\x20\x7f\xee":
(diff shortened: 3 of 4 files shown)


pgjones:
Thanks. I've added testing and validation on the outgoing and testing for the incoming. Looks as if the Python stdlib accepts [Latin-1](https://github.com/python/cpython/blob/master/Lib/http/server.py#L282) as does [gunicorn](https://github.com/benoitc/gunicorn/blob/master/gunicorn/_compat.py#L76). I've re-read RFC 3986 and I think `pchar` is the relevant abnf (should I call it that in the _abnf file?) and is defined as:
```
pchar         = unreserved / pct-encoded / sub-delims / ":" / "@"
unreserved    = ALPHA / DIGIT / "-" / "." / "_" / "~"
sub-delims    = "!" / "$" / "&" / "'" / "(" / ")"
                 / "*" / "+" / "," / ";" / "="
```
so I think latin-1 is not the spec correct thing to do. Not sure if I should change though.

h11/_events.py
```diff
@@ -55,6 +58,10 @@ def __init__(self, **kwargs):
                 if not isinstance(self.status_code, int):
                     raise LocalProtocolError("status code must be integer")
 
+        if "target" in self.__dict__:
+            validate(request_target_re, self.__dict__["target"],
```
njsmith: You can just write `self.target` here, no need to use `__dict__`.

Also, this code should move to `Request._validate` – `_EventBundle.__init__` is where we put code that runs on multiple different event types, while fields that only show up in one event type get validated in that type's `_validate` method. (This is definitely confusing, sorry.)

h11/tests/test_events.py
```diff
@@ -94,6 +94,13 @@ def test_events():
             headers=[("Host", "a"), ("Foo", "asd\x01\x02\x7f")],
             http_version="1.0")
 
+    # Request target is validated
+    for bad_char in b"\x20\x7f\xee":
```
njsmith: Can you add `\x00` to the list of test characters? That's the one most likely to mess people up if we let it through.

h11/tests/test_io.py
```diff
@@ -388,6 +388,14 @@ def test_reject_garbage_in_header_line():
            b"Host: foo\x00bar\r\n\r\n",
            None)
 
+def test_reject_non_vchar_in_path():
+    for bad_char in b"\x20\x7f\xee":
```
njsmith: Same comment here.

sigmavirus24:
> For outgoing requests: I would certainly hope that no one depends on sending raw utf8 or anything like that – that's what url-encoding is for. Though I suppose it's possible that there's some server out there that requires non-ascii URLs.

Requests has absolutely had bug reports because servers returned non-ascii URLs. Further, I've heard of servers putting raw utf8 characters into headers where the header value is JSON. Not to say that you have to support those, but perhaps there needs to be an escape hatch for them?

njsmith:
@sigmavirus24 This PR is specifically about the "target" field in a request (basically the request path, the thing that comes after `GET` on the request line). So it's only ever generated by the client and parsed by the server.

Headers are a bit different: for headers, the spec actually requires *allowing* bytes outside the ascii range and only forbids the ascii "control characters" (stuff like NUL and DEL and form-feed). And then as we discovered in #57, browsers actually allow control characters too, and popular software relies on this, so h11 ended up only forbidding NUL and exotic whitespace (#68). So utf8 in `Location:` or JSON values is fine.

The cases where this PR could cause a problem are:

* writing a server that needs to interoperate with a client that sends invalid binary characters (like utf-8) in the target field, without doing proper %-encoding

* writing a client that needs to interoperate with a server that insists on getting raw binary characters (like utf8) in the target field, and won't accept the proper %-encoded equivalent

It's entirely possible that something like this will come up of course, because, well, software.

Does requests provide a way to disable %-encoding and pass through invalid characters as part of the target? Have people asked for such a feature?

pgjones:
@njsmith Regarding latin-1, yes I was wondering if h11 should follow those other examples. Although I actually prefer h11 for the validation it does and think this is the correct thing to do.

I've updated to remove the nitpicks.

pgjones:
I think I understand the latin-1 usage now, [PEP 3333](https://www.python.org/dev/peps/pep-3333/#unicode-issues) passes on the encoding issues to the application using latin-1.

njsmith:
OK, code looks good and my comment above has a :+1: from @sigmavirus24, so I guess this is good to go – thanks!

PR #74, opened 2018-08-31
improve Trio server example for practical use

belm0:
* http_serve accepts request handler and debug option
  * expose receive-body code as a generator

I know it's just an example, but the Trio server code works fairly well.  With these small changes it can be imported and used in a project as-is.


smurfix:
Looks good to me 👍

njsmith:
I'm sorry, I'm not really comfortable trying to make the example server useful for practical use. To actually recommend a server for practical use we need a lot more than this – including at least tests, docs, versioning and distribution, a system for collecting bug reports, ... If this works for you and you're happy with carrying a local fork then that's awesome, I'm glad it helped. But if we're going to start trying to maintain this as an "upstream" that people push improvements back to, then that will become an ongoing thing, and we need to figure out how to handle it in a sustainable way. Recommending people copy the example server into their project and start hacking on it isn't really that.

So... let's focus on this as a pedagogical example. Does this help people understand how to use the h11 API? Bringing `async_generator` in here seems like it's asking a *lot* from readers... even native async generators often throw people for a loop.

Issue #72, opened 2018-08-20
h11.Response() doesn't honor IntEnum status_code 


belm0:
e.g. `h11.Response(status_code=HTTPStatus.NOT_FOUND)` yields a 200 status

it appears that `str()` is being used on the value

https://github.com/python-hyper/h11/blob/5a976c938204be494b243d2653a70b28fae17b58/h11/_writers.py#L54

njsmith:
Wow, that's tricky.

We already validate that `isinstance(status_code, int)`... but of course that doesn't help.

The official safe way to coerce to int is `operator.index`... but that passes through `HTTPStatus` objects unchanged.

At least `int(HTTPStatus.NOT_FOUND)` actually does the right thing.

I guess we should do `isinstance` *and* coerce using `int`? (In `_EventBundle.__init__`)

belm0:
would `("%d" % response.status_code).encode("ascii")` in the cited code be enough?

njsmith:
That seems reasonable too. I guess the main difference is whether anyone looking at `event.status_code` in the mean time sees an `int` or an `HTTPStatus`. I'm a little nervous about having a non-`int` object there since all the code was written on the assumption that it's an `int`, but of course `IntEnum` is compatible with `int` so it should be fine right? ...maybe?

belm0:
I'm tending to agree that the library should covert at entry-- it seems too error prone for something in the implementation to accidentally start using str(), `%s`, etc. on one of these objects.

The implication for library authors isn't good:  basically every API accepting integer codes needs to be defensive like this.  IntEnum is the closest thing to a grenade I've seen in Python.

shaolo1:
I actually tripped over this recently using quart [](https://github.com/pgjones/quart/issues/29)  I didn't realize at the time I submitted the issue that it was h11

njsmith:
See #77 for a fix.

PR #77, opened 2018-10-30
Force status codes to be integers, not IntEnums

njsmith:
IntEnum objects, like http.HTTPStatus objects, are subclasses of int,
but don't behave the same in all circumstances, e.g. string
formatting. Force our status codes to be actual int objects, so that
they'll behave the way we expect.

Fixes gh-72.

### DIFF
--- a/h11/_events.py
+++ b/h11/_events.py
@@ -57,6 +57,10 @@ class _EventBundle(object):
             if "status_code" in self.__dict__:
                 if not isinstance(self.status_code, int):
                     raise LocalProtocolError("status code must be integer")
+                # Because IntEnum objects are instances of int, but aren't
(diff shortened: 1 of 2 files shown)


pgjones:
This looks good to me, should we merge before #82?

njsmith:
@pgjones Oh yeah, we should do that. And add a release note.

Do you have write access? Do you want it?

pgjones:
I don't, yea I'm happy to. To be clear I'd merge this, #82 and then release on PyPI (assuming you mean that as well).

pgjones:
#82 is ready (although needs a correct release date) - I'm happy to do so with PyPI permissions, otherwise could you? (I included a variation of #81 as well)

pgjones:
This has been merged manually.

Issue #80, opened 2019-04-15
Add the current state when raising LocalProtocolError("not in a reusable state")


alanbchristie:
I'm experiencing `LocalProtocolError("not in a reusable state")`exceptions when using **hypercorn** (see https://github.com/pgjones/hypercorn/issues/18).

It would be extremely valuable if the current state were logged with the exception that occurs in **_state.py**'s `start_next_cycle(self)` method. Then the user would know what the state is as well as what it is expected to be.

njsmith:
That would make that error message more informative, wouldn't it!

Any interest in putting together a PR? The code that raises that exception is at the bottom of [h11/_state.py](https://github.com/python-hyper/h11/blob/836d95d1c2af2f9153c86dbc8d9784341d73c6a6/h11/_state.py#L309-L310).

alanbchristie:
Can do - but my 'day job' gets in the way during _normal hours_. Happy to look at this after 6pm tonight.

alanbchristie:
Changes tested and present in https://github.com/python-hyper/h11/pull/81

PR #81, opened 2019-04-15
- Patch for issue #80

alanbchristie:
This change adds the current state when raising LocalProtocolError("not in a reusable state"), which significantly simplifies the debugging of the cause of the exception.


pgjones:
Added in dc3e462e62ddc8ebe1713f3d91cc68e96054dea5 (slightly differently)

Issue #86, opened 2019-08-22
Tests fail with pytest 5


scarabeusiv:
```
[    4s] >       assert "kwarg a" in str(exc)
[    4s] E       AssertionError: assert 'kwarg a' in '<ExceptionInfo TypeError tblen=2>'
[    4s] E        +  where '<ExceptionInfo TypeError tblen=2>' = str(<ExceptionInfo TypeError tblen=2>)
[    4s] 
[    4s] h11/tests/test_events.py:47: AssertionError
```

The syntax bit changed and you just can't do exception but instead have to call exception.value,

pquentin:
Hi! Thanks for reporting that issue.

So that's an interesting situation:

* h11 can't use pytest 5 because it supports Python 2, so we actually need to specify `pytest<5` in the test requirements.
* However, I believe exc.value is the recommended form in pytest 4 too, so fixing that would be great

Would you be interested in putting together a pull request with those two changes?

Thanks

scarabeusiv:
You can use pytest 5 for python3 only, even in travis; in short it will pick pytest 4 on python2 as there is nothing newer and pytest 5.1.1 (atm) for the python3 always keeping it on the latest release.

There might be minor issue if the two start to diverge too much and ie using pytest 5.2 would break up the behaviour with pytest 4 completely, but I suppose the upstream pytest devs are not that crazy :)

pquentin:
That's an option too, yes

njsmith:
Fixed by #87

PR #87, opened 2019-08-23
Get value from exception to fix pytest 5 compat

scarabeusiv:
Let us see if oneliner is good enough. Issue #86


njsmith:
Codecov is still being flaky... I had to force it to report the results, because it was stuck waiting forever for some uploads that timed out.

Actual change looks good though :-)

PR #89, opened 2019-09-11
Add GitHub Actions CI/CD Workflow

webknjaz:


webknjaz:
Demo: https://github.com/webknjaz/h11/runs/219198825

pquentin:
Oh, interesting! Daily runs are a great feature.

webknjaz:
Yep! Plus a number of extra free envs and up to 20 parallel jobs.

njsmith:
Nice work!

I guess the main limitation is that this doesn't cover the pypy/pypy3/py38-dev tests that we run on Travis now, so we'd presumably have to keep both running? Or is there a way to cover those too? Testing on Windows and macOS is nice just in case, though the full set of OSes here seems like a bit of overkill, given that h11 is pure python and doesn't use any OS interfaces at all.

webknjaz:
The action that installs Pythons only supports those versions. Theoretically, it is possible to install pypy and dev builds but it'd require somebody to implement that. Maybe I'll create actions for those if I have time but not now. So just keep Travis CI for now and enjoy more free resources :)
The matrix can be made smaller but given the duration of test activities and the parallelism, I'd not care about it right now. Also, even though there's no OS-specific stuff in code, it's nice to test against different build configurations since those interpreters aren't compiled the same way as in other CIs.

Oh, and one more thing: codecov gets confused because the service doesn't know how to verify things going on in Actions. So you'd need to create a secret with the codecov token @ https://github.com/python-hyper/h11/settings/secrets and than put that into the workflow.

njsmith:
So if we merge this, will that activate it, or is there a separate activation step someone has to do?

For the codecov token, is there somewhere in particular we're supposed to stick that secret? Not that I'm terribly worried about codecov tokens becoming public, but might as well figure out how this is supposed to work...

webknjaz:
@pquentin @njsmith here's how it looks in the fork: https://github.com/webknjaz/h11/actions/runs/59303498

pgjones:
Closed in favour of #108 (to match other Hyper projects). Thanks for this though, it was used as the basis for what is now the Hyper workflow.

PR #94, opened 2020-01-09
Normalize Content-Length into a single value if duplicated exactly

### LINKED ISSUE #92: Handle multiple Content-Length headers with the same value
From RFC 7230 Section 3.3.2:

```
If a message is received that has multiple Content-Length header
   fields with field-values consisting of the same decimal value, or a
   single Content-Length header field with a field value containing a
   list of identical decimal values (e.g., "Content-Length: 42, 42"),
   indicating that duplicate Content-Length header fields have been
   generated or combined by an upstream message processor, then the
   recipient MUST either reject the message as invalid or replace the
   duplicated field-values with a single valid Content-Length field
   containing that decimal value prior to determining the message body
   length or forwarding the message.
```

Currently h11 will error out if multiple `Content-Length` headers are sent even when equal in value. Perhaps we should collapse the multiple headers into only one or allow the duplicate to pass through if the values are the same?

StephenBrown2:
Fixes #92

### DIFF
--- a/h11/_headers.py
+++ b/h11/_headers.py
@@ -64,7 +64,7 @@ _field_value_re = re.compile(field_value.encode("ascii"))
 
 def normalize_and_validate(headers, _parsed=False):
     new_headers = []
-    saw_content_length = False
+    saw_content_length = None
     saw_transfer_encoding = False
     for name, value in headers:
         # For headers coming out of the parser, we can safely skip some steps,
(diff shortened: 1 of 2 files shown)


h11/_headers.py
```diff
@@ -77,10 +77,13 @@ def normalize_and_validate(headers, _parsed=False):
             validate(_field_value_re, value, "Illegal header value {!r}", value)
         name = name.lower()
         if name == b"content-length":
-            if saw_content_length:
-                raise LocalProtocolError("multiple Content-Length headers")
             validate(_content_length_re, value, "bad Content-Length")
-            saw_content_length = True
+            if saw_content_length:
```
sethmlarson: We'd also like to handle `Content-Length: x, x, x, ...` -> `Content-Length: x`. cc @njsmith

pgjones: Should it also be `if saw_content_length is not None`, say if one header is `Content-Length: 0`?

pgjones:
I've followed this up with #109. Thoughts and comments desired there please.

PR #99, opened 2020-03-20
Make error messages returned by match failures less ambiguous (#98)

### LINKED ISSUE #98: RemoteProtocolError's messaging is not specific
There are several conditions that lead to a "malformed data" RemoteProtocolError. This makes sending specific messaging to the user (such as "Illegal header line.", etc) difficult.

Is there a way that we can gather more information about why this particular message occurs?

ian-otto:
I have updated the messages that h11 provides in Local/RemoteProtocolError's. I attempted to somewhat follow the way the other messages were formatted, but I would be open to making any changes that would be desired.

Resolves #98.


webknjaz:
Plz update the PR title to make it more informative.

webknjaz:
Use `Resolves #98.` in the text, GitHub doesn't know the "solves" keyword.

duper51:
@webknjaz Updated title and changed wording in the PR

njsmith:
Looks basically good to me. Should we also enhance `validate` to include the offending data in the error message?

duper51:
I was curious about that, I wasn't sure it would be necessary. The user would theoretically already have access to anything we would return to them with the new messages. 

It would be useful to know where the parser was tripping up, but I'm unsure of how to implement that since (in an example of a request-line validation failure) the whole line is parsed by a singular regex match.

njsmith:
> I was curious about that, I wasn't sure it would be necessary. The user would theoretically already have access to anything we would return to them with the new messages.

In theory the user can figure out anything, but it's always nice when error messages give you the relevant information directly :-)

> It would be useful to know where the parser was tripping up, but I'm unsure of how to implement that since (in an example of a request-line validation failure) the whole line is parsed by a singular regex match.

Yeah, to give fine-grained parser errors I think we'd have to move the parsing logic into Python instead of relying on the re module. And that would produce a huge slowdown, so I don't think it's worth it.

duper51:
So are you thinking we should just put the data that failed to parse after the failure message (e.g "invalid header line: garbage input\r\n")?

njsmith:
Yeah, something like that.

duper51:
I have bumped this in.

pquentin:
(I have a vague feeling that including arbitrarily large user input in an exception value is a bad idea, but I can't think of any actual vulnerability.)

duper51:
Perhaps we can cut the text after a certain length? I agree. At the very least it could be a DoS given that we're effectively doubling the memory allocation for whatever they give.

webknjaz:
> At the very least it could be a DoS given that we're effectively doubling the memory allocation for whatever they give.

Do you imply that strings are copied rather than being passed by reference?

duper51:
When appended to like `"some string: " + message`, it is my understanding that the data contained in the message will be duplicated in memory, one with "some string: " prepended, and one is the original.

That is correct, right?

webknjaz:
@duper51 I think so but I'd worry about it later. It's a little bit premature atm.

webknjaz:
@pquentin @njsmith what does this PR need to move forward?

webknjaz:
@njsmith @pquentin is it possible to merge this PR?

pquentin:
It looks good to me but I don't have merge rights :)

ian-otto:
It probably requires a rebase, I'll get that done and have this ready to be put in.

webknjaz:
@njsmith do you think you could accept this now?

webknjaz:
@njsmith anything preventing the release?

Issue #31, opened 2017-02-19
Discussion: Capitalize the header names of outgoing headers.


futursolo:
Although, as per RFC 2616, All the HTTP/1.1 headers should be case-insensitive. However, as a general practice, most of the major web servers have their headers capitalized. Also, some broken HTTP/1.1 implementations do not recognize some specific headers(e.g.: Server header for Apache Benchmark) if they are not properly capitalized. It would be better if the header can be capitalized.

Performance Concerns: Compared to `str.lower()`, str.title() is slower. However, this can be migrated by adding an LRU Cache into the header capitalization process to solve the problem.

Lukasa:
I strongly recommend making it possible to emit headers with any capitalisation, while making the current behaviour the default. 

The reality is that HTTP implementations that require title cased headers are wrong. Their inability to tolerate them, forcing other implementations to psychically tolerate their preferences, is a bad behaviour and should be punished. However, it should clearly be *possible* to overrule that behaviour if needed to make an application work. 

Note that the "capitalise as a matter of course" servers and clients are on the way out. HTTP/2 forcibly lowercases header names, which means a lot of these bad implementations get fixed as they get support for HTTP/2. So this may be a problem that reduces in scope over time. 

A final note: RFC 2616 is dead. Long live RFCs 723[0-3]!

sigmavirus24:
ITYM 723[0-5]

njsmith:
I doubt there's any real performance difference either way between titlecasing and lowercasing, at least at the individual string level:
```
~$ pyperf timeit 'b"content-length".lower()'                            
.....................
Median +- std dev: 67.3 ns +- 2.3 ns
~$ pyperf timeit 'b"content-length".title()'
.....................
Median +- std dev: 75.2 ns +- 4.0 ns
```
The difference here is far smaller than the overhead of any caching scheme we could come up with.

In general I'd strongly prefer not to have a configuration knob here if at all possible, due to the costs to devs and users of maintaining, testing, documenting the thing...

[...]

PR #103, opened 2020-08-25
Preserve header casing

tomchristie:
A proposal for preserving header casing information in both directions, while still exposing lower-case only to users by default, and keeping careful type separation. 

Essentially a take on @Lukasa's comment here https://github.com/python-hyper/h11/issues/31#issuecomment-281013856 and @njsmith's follow up...

> write a data structure that is basically a tuple, but allows you to access headers either case sensitively or case insensitively (where case insensitively is the default)

> Here's a really simple (maybe too simple) idea to throw into the mix: use tuples of (header, header_with_original_casing, value)

Here's how it looks to the end-user...

* `h11` now preserves header casing on sending headers.
* `.headers` becomes a `Sequence` type, rather than a raw list. Iterating over the sequence continues to return `(<lowercased-name>, <value>)` pairs.
* `.headers.raw()` is available for usages such as console or debug output that require original casing information, and returns a list of `(<lowercased-name>, <raw-name>, <value>)` three-tuples.

I've addressed this commit-by-commit, which should help make the approach I've taken here clear...

1. Headers becomes a `Sequence` type, rather than a raw mutable list, but continues to store exactly the same information.
2. Store raw casing information in the `Headers` type, but don't use it anywhere.
3. Use title casing anywhere we're using `get_comma_header`, `set_comma_header`. Both functions continue to be case insensitive in their effects, but it will matter in the `set_comma_header` case because it'll give us nice header casing on the over-the-wire bytes. Switching both over within the codebase for consistency.
4. Switch the writer to use the raw header casing.

Strictly speaking there *is* an API change here, in that `.headers` on events are now sequences, rather than plain lists. Any user code that is doing grungy stuff by mutating that data-structure in-place, wouldn't function after this. But that's a bit of a hacky broken thing to be doing anyway, so a version bump that tightened up the API spec into "`.headers` is an immutable sequence" seems reasonable enough right?

Anyway's putting this out there, so we've got something to discuss. 🤔

Thanks so much for maintaining such a fantastically careful & thoroughly designed library. It's a joy to work with. ✨

### DIFF
--- a/h11/_headers.py
+++ b/h11/_headers.py
@@ -1,4 +1,9 @@
-from collections.abc import Sequence
+try:
+    from collections.abc import Sequence
+except ImportError:
+    # Python 2.7 support
+    from collections import Sequence
+
 import re
 
 from ._abnf import field_name, field_value
--- a/h11/_headers.py
+++ b/h11/_headers.py
@@ -73,7 +73,7 @@ class Headers(Sequence):
         self._items = items
 
     def __getitem__(self, item):
-        _, _, value = self._data[item]
+        _, _, value = self._items[item]
         return value
 
     def __len__(self):
--- a/h11/_headers.py
+++ b/h11/_headers.py
@@ -1,3 +1,4 @@
+from collections.abc import Sequence
 import re
 
 from ._abnf import field_name, field_value
@@ -62,6 +63,25 @@ _field_name_re = re.compile(field_name.encode("ascii"))
 _field_value_re = re.compile(field_value.encode("ascii"))
 
 
+class Headers(Sequence):
+    def __init__(self, items):
+        self._items = items
+
+    def __getitem__(self, item):
+        _, value = self._data[item]
+        return value
+
+    def __len__(self):
+        return len(self._items)
+
+    def __iter__(self):
+        for name, value in self._items:
+            yield name, value
+
+    def __eq__(self, other):
+        return list(self) == other
+
+
 def normalize_and_validate(headers, _parsed=False):
     new_headers = []
     saw_content_length = False
@@ -100,7 +120,7 @@ def normalize_and_validate(headers, _parsed=False):
                 )
             saw_transfer_encoding = True
         new_headers.append((name, value))
-    return new_headers
+    return Headers(new_headers)
 
 
 def get_comma_header(headers, name):
@@ -158,7 +178,7 @@ def set_comma_header(headers, name, new_values):
             new_headers.append((found_name, found_raw_value))
     for new_value in new_values:
         new_headers.append((name, new_value))
-    headers[:] = normalize_and_validate(new_headers)
+    return normalize_and_validate(new_headers)
 
 
 def has_expect_100_continue(request):
--- a/h11/_headers.py
+++ b/h11/_headers.py
@@ -1,9 +1,3 @@
-try:
-    from collections.abc import Sequence
-except ImportError:
-    # Python 2.7 support
-    from collections import Sequence
-
 import re
 
 from ._abnf import field_name, field_value
@@ -68,26 +62,18 @@ _field_name_re = re.compile(field_name.encode("ascii"))
 _field_value_re = re.compile(field_value.encode("ascii"))
 
 
-class Headers(Sequence):
-    def __init__(self, items):
-        self._items = items
-
-    def __getitem__(self, item):
-        _, _, value = self._items[item]
-        return value
-
-    def __len__(self):
-        return len(self._items)
+class Header:
+    def __init__(self, name, value):
+        self.raw_name = name
+        self.name = name.lower()
+        self.value = value
 
     def __iter__(self):
-        for name, _, value in self._items:
-            yield name, value
+        yield self.name
+        yield self.value
 
     def __eq__(self, other):
-        return list(self) == other
-
-    def raw(self):
-        return list(self._items)
+        return (self.name, self.value) == other
 
 
 def normalize_and_validate(headers, _parsed=False):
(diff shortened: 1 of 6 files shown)


h11/_headers.py
```diff
@@ -62,6 +68,28 @@
...
+        return len(self._items)
+
+    def __iter__(self):
+        for name, _, value in self._items:
+            yield name, value
+
+    def __eq__(self, other):
+        return list(self) == other
+
+    def raw(self):
```
sigmavirus24, line 89: I'm curious about this. It seems it would return a list of the triplet where elsewhere `raw` implies the name that came over the wire. This should be clearer about what it is returning

h11/_headers.py
```diff
@@ -62,6 +68,28 @@
 _field_value_re = re.compile(field_value.encode("ascii"))
 
 
+class Headers(Sequence):
+    def __init__(self, items):
+        self._items = items
+
+    def __getitem__(self, item):
+        _, _, value = self._data[item]
```
sigmavirus24, line 76: Where does `self._data` come from here?

h11/_headers.py
sigmavirus24: I also wonder about the wisdom of having a headers sequence without structured single-header data

h11/_headers.py
tomchristie: Good catch, thanks - should be `self._items`.
Missing test coverage for this method, which needs filling in.

Kriechi:
Just as a comment or maybe for additional inspiration:
This is how we solved a similar problem in the mitmproxy project:
https://github.com/mitmproxy/mitmproxy/blob/master/mitmproxy/net/http/headers.py

h11/_headers.py
tomchristie: > I also wonder about the wisdom of having a headers sequence without structured single-header data

I'm not quite sure what you mean here. Are you saying "If you're returning a three-tuple from this interface then let's have it use a named-tuple" or something else?

Perhaps a marginally different interface for us to expose here would not be `.raw() -> (<lowercase name>, <raw name>, <value>)`, but instead expose just `.raw_items() -> (<raw name>, <value>)`

Perhaps that addressed the naming/intent slightly better?

sigmavirus24: I'm thinking of having a class that encapsulates a `Header` and then having the collection of `Headers` use that. A namedtuple could work fine as well. All that said, I get that tuples may be a smidge faster and that these are internal implementation details. Speaking from having worked on header collection objects in urllib3 in the past, these tuples can drive maintainers to pull out their hair (as well as future folks trying to update/extend the behaviour).

tomchristie: Right, for something that remains compatible with the existing API I think the options here are...

* A custom `Headers` sequence, that exposes the extra information in a `.raw_items()` interface or similar, that returns a two-tuple of `(case-sensitive-name, value)` for usages that require the raw casing info.
* A custom `Headers` sequence, the returns `Header` instances, that can iterate as two-tuples, but also expose `.name`, `.case_sensitive_name` and `.value` attributes, which are available for usages that require the raw casing info.

Or some variation on those. (Eg. this PR which currently has `.raw()` returning the three-tuple of info.)
Personally I'm fairly agnostic, as both the above options seem reasonable enough. The `Header` case has the most extra overhead, since it creates and accesses a per-header instance rather than the plain tuple, while I wouldn't expect the `.raw_items()` approach to introduce anything really noticeable, but I could run through some timings on each of the options to help better inform our options.

tomchristie:
Okay, based on the feedback here's a slight retake on this, which I think exposes a neater interface and has a pretty minimal change footprint.

After this pull request...

* `h11` preserves header casing on the wire.
* `event.headers` changes marginally in that it exposes a list of `Header` instances, which are *tuple-like* (name, value) byte pairs, rather than *actual* tuples.
* The `Header` items can also be accessed explicitly, with `header.name`, `header.value`. For the case where you need the original case-sensitive naming, `header.raw_name` is also available.

Internally...

* `set_comma_header` no longer "expects name to be lowercase bytes". It keeps the same case-insensitive behaviour, while preserving whatever casing information is passed.
* `get_comma_header` no longer "expects name to be lowercase bytes". It keeps the same case-insensitive behaviour.
* Test cases use title casing whenever calling into `set_comma_header`/`get_comma_header`.

njsmith:
What do you get from `PYTHONPATH=. python bench/benchmarks/benchmarks.py` before and after this change?

tomchristie:
A thanks I'd just been looking into that, but I'd not seen the benchmark.

Just as a rough guideline, I've taken the following to get ideas of comparative performances of the different approaches here...

```python
import h11
import timeit


def send_request():
    conn = h11.Connection(our_role=h11.CLIENT)
    headers = [
        (b'Accept', b'*/*'),
        (b'Accept-Encoding', b'gzip, deflate'),
        (b'Connection', b'keep-alive'),
        (b'Host', b'www.example.org'),
        (b'User-Agent', b'HTTPie/2.2.0'),
    ]
    request = h11.Request(method="GET", target="/", headers=headers)
    bytes_to_send = conn.send(request)


print(timeit.timeit(send_request, number=100000))
```

Which comes out with...

* Plain tuples, case not preserved: ~3.5 seconds
* Headers as a sequence, exposing `.raw_items` interface: ~4.1 seconds
* List of Header instances, exposing `.raw_name` interface: ~4.5 seconds

I'll take a look at the proper benchmarks now...

tomchristie:
Existing...

```
$ PYTHONPATH=. venv/bin/python bench/benchmarks/benchmarks.py
7389.5 requests/sec
7457.9 requests/sec
7451.1 requests/sec
7445.4 requests/sec
7434.2 requests/sec
7428.8 requests/sec
7447.1 requests/sec
```

`Headers` as a sequence...

```
$ PYTHONPATH=. venv/bin/python bench/benchmarks/benchmarks.py
6393.3 requests/sec
6404.6 requests/sec
6369.1 requests/sec
6346.9 requests/sec
6372.4 requests/sec
6388.1 requests/sec
6403.9 requests/sec
```

`Header` instances...

```
$ PYTHONPATH=. venv/bin/python bench/benchmarks/benchmarks.py
5851.9 requests/sec
5897.9 requests/sec
5894.1 requests/sec
5873.7 requests/sec
5875.9 requests/sec
5871.4 requests/sec
5903.8 requests/sec
```

njsmith:
The benchmark I cited isn't terribly clever, but it does exercise a full request/response cycle with some realistic headers.

tomchristie:
Closing this off in favour of #104

PR #104, opened 2020-09-08
Preserve header casing. Take two. 🎬

### LINKED ISSUE #31: Discussion: Capitalize the header names of outgoing headers.

tomchristie:
Closes #31

Based on #103, focusing on minimising benchmark regressions.

* Preserves HTTP header casing.
* `event.headers` is a sequence-like instance, containing two-tuples of bytes.
* `event.headers.raw_items()` is a list of two-tuples of `(raw_name, value)`.

Three benchmark runs here against each case.

#### Running against current master...

7330.1 requests/sec
7357.9 requests/sec
7363.4 requests/sec
7310.5 requests/sec
7363.2 requests/sec
7354.3 requests/sec
7363.3 requests/sec

7321.4 requests/sec
7355.2 requests/sec
7353.7 requests/sec
7375.7 requests/sec
7352.6 requests/sec
7381.0 requests/sec
7410.0 requests/sec

7234.4 requests/sec
7274.0 requests/sec
7231.8 requests/sec
7272.5 requests/sec
7263.7 requests/sec
7268.9 requests/sec
7273.7 requests/sec

#### Running against this PR...

7022.7 requests/sec
7041.2 requests/sec
7052.2 requests/sec
7038.3 requests/sec
7038.5 requests/sec
7042.8 requests/sec
7053.2 requests/sec

7129.9 requests/sec
7145.6 requests/sec
7152.4 requests/sec
7158.3 requests/sec
7154.1 requests/sec
7116.6 requests/sec
7146.2 requests/sec

7080.6 requests/sec
7077.9 requests/sec
7100.3 requests/sec
7051.9 requests/sec
7091.7 requests/sec
7072.6 requests/sec
7056.3 requests/sec

### DIFF
--- a/docs/source/api.rst
+++ b/docs/source/api.rst
@@ -128,6 +128,21 @@ have a newline inside a header value, and ``Content-Length: hello`` is
 an error because `Content-Length` should always be an integer. We may
 add additional checks in the future.
 
+While we make sure to expose header names as lowercased bytes, we also
+preserve the original header casing that is used. Compliant HTTP
+agents should always treat headers in a case insensitive manner, but
+this may not always be the case. When sending bytes over the wire we
+use send headers preserving whatever original header casing is used.
+
+It is possible to access the headers in their raw original casing,
+which may be useful for some user output or debugging purposes.
+
+.. ipython:: python
+
+    original_headers = [("Host", "example.com")]
+    req = h11.Request(method="GET", target="/", headers=original_headers)
+    req.headers.raw_items()
+
 .. _http_version-format:
 
 It's not just headers we normalize to being byte-strings: the same
--- a/h11/_headers.py
+++ b/h11/_headers.py
@@ -170,14 +170,12 @@ def get_comma_header(headers, name):
 
 def set_comma_header(headers, name, new_values):
     # The header name `name` is expected to be lower-case bytes.
-    raw_name = name
-    name = name.lower()
     new_headers = []
     for found_raw_name, found_name, found_raw_value in headers.raw_items:
         if found_name != name:
             new_headers.append((found_raw_name, found_raw_value))
     for new_value in new_values:
-        new_headers.append((raw_name, new_value))
+        new_headers.append((name.title(), new_value))
     return normalize_and_validate(new_headers)
 
 
--- a/h11/_headers.py
+++ b/h11/_headers.py
@@ -103,6 +103,13 @@ class Headers:
     def __len__(self):
         return len(self._full_items)
 
+    def __repr__(self):
+        return "<Headers(%s)>" % repr(list(self))
+
+    def __getitem__(self, idx):
+         _, name, value = self._full_items[idx]
+         return (name, value)
+
     def raw_items(self):
         return [(raw_name, value) for raw_name, _, value in self._full_items]
 
--- a/h11/_headers.py
+++ b/h11/_headers.py
@@ -63,20 +63,48 @@ _field_value_re = re.compile(field_value.encode("ascii"))
 
 
 class Headers:
-    __slots__ = ['raw_items']
+    """
+    A list-like interface that allows iterating over headers as byte-pairs
+    of (lowercased-name, value).
 
-    def __init__(self, raw_items):
-        self.raw_items = raw_items
+    Internally we actually store the representation as three-tuples,
+    including both the raw original casing, in order to preserve casing
+    over-the-wire, and the lowercased name, for case-insensitive comparisions.
+
+    r = Request(
+        method="GET",
+        target="/",
+        headers=[("Host", "example.org"), ("Connection", "keep-alive")],
+        http_version="1.1",
+    )
+    assert r.headers == [
+        (b"host", b"example.org"),
+        (b"connection", b"keep-alive")
+    ]
+    assert r.headers.raw_items() == [
+        (b"Host", b"example.org"),
+        (b"Connection", b"keep-alive")
+    ]
+    """
+
+    def __init__(self, full_items):
+        self._full_items = full_items
 
     def __iter__(self):
-        for _, name, value in self.raw_items:
+        for _, name, value in self._full_items:
             yield name, value
 
     def __bool__(self):
-        return bool(self.raw_items)
+        return bool(self._full_items)
 
     def __eq__(self, other):
         return list(self) == list(other)
+
+    def __len__(self):
+        return len(self._full_items)
+
+    def raw_items(self):
+        return [(raw_name, value) for raw_name, _, value in self._full_items]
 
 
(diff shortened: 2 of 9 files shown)


tomchristie:
Okay, I think this is about the best we can do here.
I've updated the docs to include the one extra point of API that's now exposed.

Unavoidably it does benchmark marginally slower that the current `master` branch.

Before...

```
7211.1 requests/sec
7300.6 requests/sec
7298.3 requests/sec
7271.7 requests/sec
7342.2 requests/sec
7310.1 requests/sec
7334.3 requests/sec
```

After...

```
7025.6 requests/sec
7094.9 requests/sec
7095.7 requests/sec
7094.9 requests/sec
7115.5 requests/sec
7053.5 requests/sec
7104.2 requests/sec
```

Averaging to: 7295 requests/sec -> 7083 requests/sec

What do we think about this folks? @njsmith, @pgjones?

(From my POV I'd really like to get this resolved - we're currently treating it as one of our HTTPX 1.0 blockers)

docs/source/api.rst
```diff
@@ -128,6 +128,21 @@ have a newline inside a header value, and ``Content-Length: hello`` is
 an error because `Content-Length` should always be an integer. We may
 add additional checks in the future.
 
+While we make sure to expose header names as lowercased bytes, we also
+preserve the original header casing that is used. Compliant HTTP
+agents should always treat headers in a case insensitive manner, but
+this may not always be the case. When sending bytes over the wire we
+use send headers preserving whatever original header casing is used.
```
StephenBrown2, line 135: "use send headers" is confusing. Perhaps you meant?:
```suggestion
send headers preserving whatever original header casing was used.
```

docs/source/api.rst
```diff
@@ -128,6 +128,21 @@ have a newline inside a header value, and ``Content-Length: hello`` is
...
+
+It is possible to access the headers in their raw original casing,
+which may be useful for some user output or debugging purposes.
+
+.. ipython:: python
+
+    original_headers = [("Host", "example.com")]
+    req = h11.Request(method="GET", target="/", headers=original_headers)
+    req.headers.raw_items()
```
StephenBrown2, line 144: Will there be output demonstrating the result?

tomchristie: Yup, see https://h11.readthedocs.io/en/latest/api.html - but I wasn't able to figure out the docs build locally just yet, so I've not seen the rendered docs myself.

pquentin: You can see the docs rendered for this pull request at https://h11--104.org.readthedocs.build/en/104/

tomchristie: Ah neat, thanks @pquentin.

Not 100% obvious what repr we'd want for the `Headers` there:

* Use `<Headers([(b'Host', b'example.org')])>` to make it clear it's not (quite) a plain list.
* Use `[(b'Host', b'example.org')]` to keep it looking simple.
* Use `((b'Host', b'example.org'),)` since it's a non-mutable sequence.
* Something else?

Some examples [as currently rendered](https://h11--104.org.readthedocs.build/en/104/api.html)...

Screen Shot 2020-09-15 at 10 37 06

Screen Shot 2020-09-15 at 10 37 39

pquentin:
@StephenBrown2 @pgjones Would one of you be able to review this fine work? I'm not qualified (and not part of python-hyper anyways) and @njsmith isn't currently available for OSS reviews (even on Trio)

h11/_headers.py
```diff
@@ -153,12 +207,12 @@ def get_comma_header(headers, name):
...
     new_headers = []
-    for found_name, found_raw_value in headers:
+    for found_raw_name, found_name, found_raw_value in headers._full_items:
         if found_name != name:
-            new_headers.append((found_name, found_raw_value))
+            new_headers.append((found_raw_name, found_raw_value))
     for new_value in new_values:
-        new_headers.append((name, new_value))
-    headers[:] = normalize_and_validate(new_headers)
+        new_headers.append((name.title(), new_value))
```
pgjones, line 214: I think this is my only concern, as it changes the h11 set header casing to title rather than lowercase. Which could be problematic if lowercase headers are desired (no way to avoid h11 writing titlecase now). Does you usecase require titlecased response headers as well? Alternatively, would simply knowing the raw casing be sufficient?

tomchristie: So, `get_comma_header` and `set_comma_header` both expect lowercase headers.

We've got two options with this PR:

* Switch both `get_comma_header` and `set_comma_header` to expect case-sensitive headers, and lowercase them for the lookup behaviour. Preserve whatever header casing is used when calling `set_comma_header`.
* Continue to expect lowercase headers to both `get_comma_header` and `set_comma_header`. When setting up the header casing in `set_comma_header`, use title casing.

Of those two, I've gone for the second, as being marginally less impactful, and also being marginally quicker. (Only requires a casing transform in one of the two cases.)

Both of these two are private methods, so we're okay for them to have behaviour that only makes sense if they're being used in an appropriate context. In particular, the docstring for `get_comma_header` already lists the full set of header cases that it's actually being used with.

I've now added a corresponding docstring to `set_comma_header`, which I think helps clarify (internally), why the title casing is sufficient and appropriate for this function.

pgjones: I don't think I explained my concern well enough, however you have given answer I was seeking 

> we use title casing for the header names, in order to match the conventional HTTP header style.

For clarity my concern is that this choice means that I can't use h11 to exclusively write lowercase headers over the wire (equally with the current version I can't use h11 to exclusively write titlecased headers). 

I think this change is ok - I'll note it in the release notes and we can always add another flag if someone really needs to only write lowercase headers.

pgjones (approved):

pgjones:
Minded to merge - happy to be corrected - will merge and release over the weekend if not.

h11/_headers.py
sigmavirus24: I feel like I've seen cases where people complain about using lowercase'd headers because the server they were talking to didn't support them. Title cased is probably (sadly) marginally safer. I think preserving user-casing is probably the least surprising but also most expensive

tomchristie:
@pgjones Fantastic - pretty excited to get this one resolved. 🤗
Presumably we'll bump to 0.11 for this release?

PR #109, opened 2020-10-04
Content length handling

### LINKED ISSUE #92: Handle multiple Content-Length headers with the same value

pgjones:
Fixes #92

### DIFF
--- a/h11/_headers.py
+++ b/h11/_headers.py
@@ -118,7 +118,7 @@ class Headers:
 
 def normalize_and_validate(headers, _parsed=False):
     new_headers = []
-    saw_content_length = None
+    seen_content_length = None
     saw_transfer_encoding = False
     for name, value in headers:
         # For headers coming out of the parser, we can safely skip some steps,
@@ -133,13 +133,13 @@ def normalize_and_validate(headers, _parsed=False):
         name = name.lower()
         if name == b"content-length":
             validate(_content_length_re, value, "bad Content-Length")
-            if saw_content_length:
-                if saw_content_length != value:
+            if seen_content_length is None:
+                seen_content_length = value
+                new_headers.append((raw_name, name, value))
+            else:
+                if seen_content_length != value:
                     raise LocalProtocolError("conflicting Content-Length headers")
-                # We have a duplicate Content-Length, so collapse it
-                new_headers.remove((name, saw_content_length))
-            saw_content_length = value
-        if name == b"transfer-encoding":
+        elif name == b"transfer-encoding":
             # "A server that receives a request message with a transfer coding
             # it does not understand SHOULD respond with 501 (Not
             # Implemented)."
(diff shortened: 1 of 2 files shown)


pgjones:
@tomchristie & @sethmlarson @StephenBrown2 Does this look good to you? I think it is probably worth getting into the next release.

h11/_headers.py
```diff
@@ -132,11 +132,19 @@ def normalize_and_validate(headers, _parsed=False):
         raw_name = name
         name = name.lower()
         if name == b"content-length":
-            if saw_content_length:
-                raise LocalProtocolError("multiple Content-Length headers")
+            if b"," in value:
+                lengths = set(length.strip() for length in value.split(b","))
+                if len(lengths) != 1:
+                    raise LocalProtocolError("conflicting Content-Length headers")
+                value = list(lengths)[0]
```
sethmlarson, line 139: You can use `lengths.pop()` here.

h11/_headers.py
```diff
@@ -132,11 +132,19 @@ def normalize_and_validate(headers, _parsed=False):
         raw_name = name
         name = name.lower()
         if name == b"content-length":
-            if saw_content_length:
-                raise LocalProtocolError("multiple Content-Length headers")
+            if b"," in value:
```
sethmlarson, line 135: Since `b"," in value` already does a full run of the string does it make sense to call `.split(b",")` then the set logic if `len(...) > 1`?

h11/_headers.py
```diff
@@ -132,11 +132,19 @@ def normalize_and_validate(headers, _parsed=False):
...
+                    raise LocalProtocolError("conflicting Content-Length headers")
+                value = list(lengths)[0]
             validate(_content_length_re, value, "bad Content-Length")
-            saw_content_length = True
-        if name == b"transfer-encoding":
+            if seen_content_length is None:
+                seen_content_length = value
+                new_headers.append((raw_name, name, value))
+            else:
+                if seen_content_length != value:
```
sethmlarson, line 145: This can be an `elif`?

sethmlarson (changes requested):
Some performance optimizations, looks good otherwise.

h11/_headers.py
pgjones: Thanks, changed. (Originally though list(...)[0] was clearer, but it wasn't strongly held)

h11/_headers.py
pgjones: Yep, changed. Thanks.

h11/_headers.py
pgjones: I think so, I've changed it. Thanks

StephenBrown2 (approved):
Seems good to me.

sethmlarson (approved):
LGTM, thanks for this!

Issue #7, opened 2016-06-04
Support for servers with broken line endings


njsmith:
I'm informed that some allegedly-HTTP servers use `\r` or `\n` for line endings, and that robust clients need to support this too.

Sigh.

Maybe the thing to do is to watch for `\r\r` or `\n\n` in the headers block and then go into a special case mode if we see it?

njsmith:
(NB I'm probably not going to work on this any time soon)

njsmith:
CC: @durin42 since it's his fault I know about this

Lukasa:
From RFC 7230 Section 3.5 (Message Parsing Robustness):

> Although the line terminator for the start-line and header fields is the sequence CRLF, a recipient MAY recognize a single LF as a line terminator and ignore any preceding CR.

This suggests that allowing simply `\n` and not `\r` is acceptable, which I think is probably the way to go. While *technically* just `'\r` *might* happen, I really don't think it's worth it in terms of compat cost.

durin42:
> On Jun 4, 2016, at 4:06 PM, Cory Benfield <notifications@github.com> wrote:
> 
> This suggests that allowing simply \n and not \r is acceptable, which I think is probably the way to go. While technically just '\r might happen, I really don't think it's worth it in terms of compat cost.
> 
> 

Sadly, I’ve *seen* \r without \n in the wild. :(

Lukasa:
Oh, so have I. But I am a strong advocate of telling implementations that haven't read RFCs to go take a long walk off a short pier.

njsmith:
I'm not sure it makes much difference in terms of implementation simplicity, because h11's trick for simple fast HTTP header parsing is use `bytearray.find("\r\n\r\n")`, so we'd still have to duplicate this check to look for `\n\n` at least.

BTW, I assume that servers that screw this up for headers also screw it up for chunked encoding?

Lukasa:
@njsmith No need to duplicate the check: better to swap to a fairly simple regex. Either way you desperately need to avoid having too much cleverness here in Python, because each time you come back into Python code to process the bytestream you slow down a *load* on CPython (which for better or worse is what users will use to judge your perf).

njsmith:
[...]

PR #115, opened 2020-11-06
Added ability to use LF, not only CRLF delimiter for response Headers and Body

cdeler:
Hello,

I want to submit a PR which closes https://github.com/python-hyper/h11/issues/7

#### Why the changes are required
According to [this](https://github.com/python-hyper/h11/issues/7#issuecomment-722464047) comment in the issue, there are problems with some old servers, which are not totally fit `HTTP/1.1` RFC. The original issue in `httpx` (https://github.com/encode/httpx/issues/1378) describes the situation, where some embedded system developers have to deal with a non-standard server

#### What has been done?
I tried to reimplement the function, which extracts headers for response, using regex

#### What hasn't done yet
~Performance testing, fuzzing~

### _(updates)_ How `maybe_extract_lines` works for now?
1. first of all it extracts part of `self._data` buffer until the `"\n\r?\n"`, which gives `data`
1. then it determines the line delimiter in `data` using `"\r?\n"` regex, which gives `delimiter`
1. then split lines using bytearray's `.split(delimiter)` (it much faster than `regex.split(data)`)

#### With fix
```
6099.7 requests/sec
6172.2 requests/sec
6094.7 requests/sec
6202.1 requests/sec
6303.3 requests/sec
6358.0 requests/sec
6372.2 requests/sec
```

#### Without fix (522b004)
```
6065.5 requests/sec
6211.9 requests/sec
6211.1 requests/sec
6131.1 requests/sec
6333.1 requests/sec
6103.8 requests/sec
6260.0 requests/sec
```

### DIFF
--- a/h11/_receivebuffer.py
+++ b/h11/_receivebuffer.py
@@ -5,7 +5,8 @@ __all__ = ["ReceiveBuffer"]
 
 
 # Operations we want to support:
-# - find next \r\n or \r\n\r\n, or wait until there is one
+# - find next \r\n or \r\n\r\n (\n or \n\n are also acceptable),
+#   or wait until there is one
 # - read at-most-N bytes
 # Goals:
 # - on average, do this fast
@@ -40,7 +41,7 @@ __all__ = ["ReceiveBuffer"]
 # processed a whole event, which could in theory be slightly more efficient
 # than the internal bytearray support.)
 
-body_and_headers_delimiter_regex = re.compile(b"\n\r?\n", re.MULTILINE)
+blank_line_delimiter_regex = re.compile(b"\n\r?\n", re.MULTILINE)
 line_delimiter_regex = re.compile(b"\r?\n", re.MULTILINE)
 
 
@@ -50,7 +51,8 @@ class ReceiveBuffer(object):
         # These are both absolute offsets into self._data:
         self._start = 0
         self._looked_at = 0
-        self._looked_for_regex = body_and_headers_delimiter_regex
+
+        self._looked_for_regex = blank_line_delimiter_regex
 
     def __bool__(self):
         return bool(len(self))
@@ -129,7 +131,7 @@ class ReceiveBuffer(object):
             self._start += len(start_chunk)
             return []
         else:
-            data = self.maybe_extract_until_next(body_and_headers_delimiter_regex, 3)
+            data = self.maybe_extract_until_next(blank_line_delimiter_regex, 3)
             if data is None:
                 return None
 
--- a/h11/_receivebuffer.py
+++ b/h11/_receivebuffer.py
@@ -41,94 +41,98 @@ __all__ = ["ReceiveBuffer"]
 # processed a whole event, which could in theory be slightly more efficient
 # than the internal bytearray support.)
 
-blank_line_delimiter_regex = re.compile(b"\n\r?\n", re.MULTILINE)
-
-
-def rstrip_line(line):
-    return line.rstrip(b"\r")
+blank_line_regex = re.compile(b"\n\r?\n", re.MULTILINE)
 
 
 class ReceiveBuffer(object):
     def __init__(self):
         self._data = bytearray()
-        # These are both absolute offsets into self._data:
-        self._start = 0
-        self._looked_at = 0
-
-        self._looked_for_regex = blank_line_delimiter_regex
-
-    def __bool__(self):
-        return bool(len(self))
-
-    # for @property unprocessed_data
-    def __bytes__(self):
-        return bytes(self._data[self._start :])
-
-    if sys.version_info[0] < 3:  # version specific: Python 2
-        __str__ = __bytes__
-        __nonzero__ = __bool__
-
-    def __len__(self):
-        return len(self._data) - self._start
-
-    def compress(self):
-        # Heuristic: only compress if it lets us reduce size by a factor
-        # of 2
-        if self._start > len(self._data) // 2:
-            del self._data[: self._start]
-            self._looked_at -= self._start
-            self._start -= self._start
+        self._next_line_search = 0
+        self._multiple_lines_search = 0
 
     def __iadd__(self, byteslike):
         self._data += byteslike
         return self
 
+    def __bool__(self):
+        return bool(len(self))
+
+    def __len__(self):
+        return len(self._data)
+
+    # for @property unprocessed_data
+    def __bytes__(self):
+        return bytes(self._data)
+
+    if sys.version_info[0] < 3:  # version specific: Python 2
+        __str__ = __bytes__
+        __nonzero__ = __bool__
+
     def maybe_extract_at_most(self, count):
-        out = self._data[self._start : self._start + count]
+        """
+        Extract a fixed number of bytes from the buffer.
+        """
+        out = self._data[:count]
         if not out:
             return None
-        self._start += len(out)
+
+        self._data[:count] = b""
+        self._next_line_search = 0
+        self._multiple_lines_search = 0
         return out
 
-    def maybe_extract_until_next(self, needle_regex, max_needle_length):
-        # Returns extracted bytes on success (advancing offset), or None on
-        # failure
-        if self._looked_for_regex == needle_regex:
-            looked_at = max(self._start, self._looked_at - max_needle_length)
-        else:
-            looked_at = self._start
-            self._looked_for_regex = needle_regex
-
-        delimiter_match = next(
-            self._looked_for_regex.finditer(self._data, looked_at), None
-        )
-
-        if delimiter_match is None:
-            self._looked_at = len(self._data)
+    def maybe_extract_next_line(self):
+        """
+        Extract the first line, if it is completed in the buffer.
+        """
+        # Only search in buffer space that we've not already looked at.
+        partial_buffer = self._data[self._next_line_search :]
+        partial_idx = partial_buffer.find(b"\n")
+        if partial_idx == -1:
+            self._next_line_search = len(self._data)
             return None
 
-        _, end = delimiter_match.span(0)
-
-        out = self._data[self._start : end]
-
-        self._start = end
-
+        # Truncate the buffer and return it.
+        idx = self._next_line_search + partial_idx + 1
+        out = self._data[:idx]
+        self._data[:idx] = b""
+        self._next_line_search = 0
+        self._multiple_lines_search = 0
         return out
 
-    # HTTP/1.1 has a number of constructs where you keep reading lines until
-    # you see a blank one. This does that, and then returns the lines.
     def maybe_extract_lines(self):
-        if self._data[self._start : self._start + 2] == b"\r\n":
-            self._start += 2
+        """
+        Extract everything up to the first blank line, and return a list of lines.
+        """
+        # Handle the case where we have an immediate empty line.
+        if self._data[:1] == b"\n":
+            self._data[:1] = b""
+            self._next_line_search = 0
+            self._multiple_lines_search = 0
             return []
-        elif self._data[self._start : self._start + 1] == b"\n":
-            self._start += 1
+
+        if self._data[:2] == b"\r\n":
+            self._data[:2] = b""
+            self._next_line_search = 0
+            self._multiple_lines_search = 0
             return []
-        else:
-            data = self.maybe_extract_until_next(blank_line_delimiter_regex, 3)
-            if data is None:
-                return None
 
-            lines = list(map(rstrip_line, data.rstrip(b"\r\n").split(b"\n")))
+        # Only search in buffer space that we've not already looked at.
+        partial_buffer = self._data[self._multiple_lines_search :]
+        match = blank_line_regex.search(partial_buffer)
+        if match is None:
+            self._multiple_lines_search = max(0, len(self._data) - 2)
+            return None
 
-            return lines
+        # Truncate the buffer and return it.
+        idx = self._multiple_lines_search + match.span(0)[-1]
+        out = self._data[:idx]
+        lines = [line.rstrip(b"\r") for line in out.split(b"\n")]
+
+        self._data[:idx] = b""
+        self._next_line_search = 0
+        self._multiple_lines_search = 0
+
+        assert lines[-2] == lines[-1] == b""
+
+        return lines[:-2]
--- a/h11/_receivebuffer.py
+++ b/h11/_receivebuffer.py
@@ -41,7 +41,7 @@ __all__ = ["ReceiveBuffer"]
 # processed a whole event, which could in theory be slightly more efficient
 # than the internal bytearray support.)
 
-blank_line_delimiter_regex = re.compile(b"(\r\n\r\n|\n\n)", re.MULTILINE)
+blank_line_delimiter_regex = re.compile(b"\n\r?\n", re.MULTILINE)
 line_delimiter_regex = re.compile(b"\r?\n", re.MULTILINE)
 
 
@@ -112,17 +112,6 @@ class ReceiveBuffer(object):
 
         return out
 
-    def _get_fields_delimiter(self, data, lines_delimiter_regex):
-        delimiter_match = next(lines_delimiter_regex.finditer(data), None)
-
-        if delimiter_match is not None:
-            begin, end = delimiter_match.span(0)
-            result = data[begin:end]
-        else:
-            result = b"\r\n"
-
-        return bytes(result)
-
     # HTTP/1.1 has a number of constructs where you keep reading lines until
     # you see a blank one. This does that, and then returns the lines.
     def maybe_extract_lines(self):
@@ -133,13 +122,10 @@ class ReceiveBuffer(object):
             self._start += 1
             return []
         else:
-            data = self.maybe_extract_until_next(blank_line_delimiter_regex, 4)
+            data = self.maybe_extract_until_next(blank_line_delimiter_regex, 3)
             if data is None:
                 return None
 
-            real_lines_delimiter = self._get_fields_delimiter(
-                data, line_delimiter_regex
-            )
-            lines = data.rstrip(b"\r\n").split(real_lines_delimiter)
+            lines = line_delimiter_regex.split(data.rstrip(b"\r\n"))
 
             return lines
--- a/h11/_receivebuffer.py
+++ b/h11/_receivebuffer.py
@@ -41,7 +41,7 @@ __all__ = ["ReceiveBuffer"]
 # processed a whole event, which could in theory be slightly more efficient
 # than the internal bytearray support.)
 
-blank_line_delimiter_regex = re.compile(b"\n\r?\n", re.MULTILINE)
+blank_line_delimiter_regex = re.compile(b"(\r\n\r\n|\n\n)", re.MULTILINE)
 line_delimiter_regex = re.compile(b"\r?\n", re.MULTILINE)
 
 
@@ -126,12 +126,14 @@ class ReceiveBuffer(object):
     # HTTP/1.1 has a number of constructs where you keep reading lines until
     # you see a blank one. This does that, and then returns the lines.
     def maybe_extract_lines(self):
-        start_chunk = self._data[self._start : self._start + 2]
-        if start_chunk in [b"\r\n", b"\n"]:
-            self._start += len(start_chunk)
+        if self._data[self._start : self._start + 2] == b"\r\n":
+            self._start += 2
+            return []
+        elif self._data[self._start : self._start + 1] == b"\n":
+            self._start += 1
             return []
         else:
-            data = self.maybe_extract_until_next(blank_line_delimiter_regex, 3)
+            data = self.maybe_extract_until_next(blank_line_delimiter_regex, 4)
             if data is None:
                 return None
 
--- a/h11/_receivebuffer.py
+++ b/h11/_receivebuffer.py
@@ -127,7 +127,11 @@ class ReceiveBuffer(object):
         # Truncate the buffer and return it.
         idx = match.span(0)[-1]
         out = self._extract(idx)
-        lines = [line.rstrip(b"\r") for line in out.split(b"\n")]
+        lines = out.split(b"\n")
+
+        for line in lines:
+            if line.endswith(b"\r"):
+                del line[-1]
 
         assert lines[-2] == lines[-1] == b""
 
--- a/h11/_receivebuffer.py
+++ b/h11/_receivebuffer.py
@@ -42,7 +42,10 @@ __all__ = ["ReceiveBuffer"]
 # than the internal bytearray support.)
 
 blank_line_delimiter_regex = re.compile(b"\n\r?\n", re.MULTILINE)
-line_delimiter_regex = re.compile(b"\r?\n", re.MULTILINE)
+
+
+def rstrip_line(line):
+    return line.rstrip(b"\r")
 
 
 class ReceiveBuffer(object):
@@ -126,6 +129,6 @@ class ReceiveBuffer(object):
             if data is None:
                 return None
 
-            lines = line_delimiter_regex.split(data.rstrip(b"\r\n"))
+            lines = list(map(rstrip_line, data.rstrip(b"\r\n").split(b"\n")))
 
             return lines
--- a/h11/_receivebuffer.py
+++ b/h11/_receivebuffer.py
@@ -95,7 +95,7 @@ class ReceiveBuffer(object):
         else:
             looked_at = self._start
             self._looked_for = needle
-            # re.compile operation is more expensive than just byte compare
+            # Check if default delimiter to avoid expensive re.compile
             if needle == default_delimiter:
                 self._looked_for_regex = delimiter_regex
             else:
--- a/h11/_receivebuffer.py
+++ b/h11/_receivebuffer.py
@@ -40,8 +40,7 @@ __all__ = ["ReceiveBuffer"]
 # processed a whole event, which could in theory be slightly more efficient
 # than the internal bytearray support.)
 
-default_delimiter = b"\n\r?\n"
-delimiter_regex = re.compile(b"\n\r?\n", re.MULTILINE)
+body_and_headers_delimiter_regex = re.compile(b"\n\r?\n", re.MULTILINE)
 line_delimiter_regex = re.compile(b"\r?\n", re.MULTILINE)
 
 
@@ -51,8 +50,7 @@ class ReceiveBuffer(object):
         # These are both absolute offsets into self._data:
         self._start = 0
         self._looked_at = 0
-        self._looked_for = default_delimiter
-        self._looked_for_regex = delimiter_regex
+        self._looked_for_regex = body_and_headers_delimiter_regex
 
     def __bool__(self):
         return bool(len(self))
@@ -87,19 +85,14 @@ class ReceiveBuffer(object):
         self._start += len(out)
         return out
 
-    def maybe_extract_until_next(self, needle):
+    def maybe_extract_until_next(self, needle_regex, max_needle_length):
         # Returns extracted bytes on success (advancing offset), or None on
         # failure
-        if self._looked_for == needle:
-            looked_at = max(self._start, self._looked_at - len(needle) + 1)
+        if self._looked_for_regex == needle_regex:
+            looked_at = max(self._start, self._looked_at - max_needle_length)
         else:
             looked_at = self._start
-            self._looked_for = needle
-            # Check if default delimiter to avoid expensive re.compile
-            if needle == default_delimiter:
-                self._looked_for_regex = delimiter_regex
-            else:
-                self._looked_for_regex = re.compile(needle, re.MULTILINE)
+            self._looked_for_regex = needle_regex
 
         delimiter_match = next(
             self._looked_for_regex.finditer(self._data, looked_at), None
@@ -136,11 +129,11 @@ class ReceiveBuffer(object):
             self._start += len(start_chunk)
             return []
         else:
-            data = self.maybe_extract_until_next(default_delimiter)
+            data = self.maybe_extract_until_next(body_and_headers_delimiter_regex, 3)
             if data is None:
                 return None
 
-            delimiter = self._get_fields_delimiter(data, line_delimiter_regex)
-            lines = data.rstrip(b"\r\n").split(delimiter)
+            real_lines_delimiter = self._get_fields_delimiter(data, line_delimiter_regex)
+            lines = data.rstrip(b"\r\n").split(real_lines_delimiter)
 
             return lines
--- a/h11/_receivebuffer.py
+++ b/h11/_receivebuffer.py
@@ -51,10 +51,8 @@ class ReceiveBuffer(object):
         # These are both absolute offsets into self._data:
         self._start = 0
         self._looked_at = 0
-        self._looked_for = b""
-
-        self._delimiter = b"\n\r?\n"
-        self._delimiter_regex = delimiter_regex
+        self._looked_for = default_delimiter
+        self._looked_for_regex = delimiter_regex
 
     def __bool__(self):
         return bool(len(self))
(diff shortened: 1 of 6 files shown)


cdeler:
### Benchmark results
#### with fix
```
(h11) h11/bench % PYTHONPATH=.. python benchmarks/benchmarks.py
5877.3 requests/sec
5874.9 requests/sec
5529.5 requests/sec
5765.3 requests/sec
5903.8 requests/sec
5624.4 requests/sec
5861.4 requests/sec
```
#### without fix (latest master 522b00470b5776dda2dc7e98ff778eb1e418e6ad)
```
(h11) h11/bench % PYTHONPATH=.. python benchmarks/benchmarks.py
6161.7 requests/sec
6201.7 requests/sec
6268.1 requests/sec
6069.3 requests/sec
6105.4 requests/sec
6256.2 requests/sec
6294.8 requests/sec
```

bdraco:
If performance is an issue, maybe look for "\r" or "\n" first.

If you see first "\r" use "\r\n" as the delimiter, otherwise use "\n".

cdeler:
@bdraco , yes, `b"test-test-test".split(b"-")` is much faster than `regex::split(...)`

cdeler:
After some performance investigations I changed the algorithm.

### How `maybe_extract_lines` works for now?

1. first of all it extracts part of `self._data` buffer until the `"\n\r?\n"`, which gives `data`
1. then it determines the line delimiter in `data` using `"\r?\n"` regex, which gives `delimiter`
1. then split lines using bytearray `.split(delimiter)` (it much faster than `regex.split(data)`)


#### With fix
```
6065.5 requests/sec
6211.9 requests/sec
6211.1 requests/sec
6131.1 requests/sec
6333.1 requests/sec
6103.8 requests/sec
6260.0 requests/sec
```

#### Without fix (522b00470b5776dda2dc7e98ff778eb1e418e6ad)
```
6065.5 requests/sec
6211.9 requests/sec
6211.1 requests/sec
6131.1 requests/sec
6333.1 requests/sec
6103.8 requests/sec
6260.0 requests/sec
```

mkanet:
@bdraco do you mind verifying the REST sensor Issue I'm having [Issue #42608](https://github.com/home-assistant/core/issues/42608) Home Assistant 0.117.x is caused by the issue described here?  I posted server response examples [here] (https://github.com/home-assistant/core/issues/42608#issuecomment-725312591).  The server response comes from the latest Glances API.

bdraco:
I'm not running glances so it's not something I'm in a position to verify.  Their would be best to checkout this PR and update the system that is experiencing the issue.

mkanet:
@bdraco thank you for responding.  I not sure how to check out this PR.  Any ideas which Home Assistant beta version will include the changes from this PR?

bdraco:
Let's keep the discussion in the Home Assistant issue as it's off topic for this PR

h11/_receivebuffer.py
```diff
@@ -83,30 +91,56 @@ def maybe_extract_until_next(self, needle):
...
-            search_start = max(self._start, self._looked_at - len(needle) + 1)
+            looked_at = max(self._start, self._looked_at - len(needle) + 1)
         else:
-            search_start = self._start
-        offset = self._data.find(needle, search_start)
-        if offset == -1:
-            self._looked_at = len(self._data)
+            looked_at = self._start
             self._looked_for = needle
+            # re.compile operation is more expensive than just byte compare
```
pgjones, line 98: If I understand this comment correctly, you are justifying the check rather than just always compiling? If so maybe say `Check if default delimiter to avoid expensive re.compile`? As I was originally confused by this comment.

h11/_receivebuffer.py
```diff
@@ -83,30 +91,56 @@ def maybe_extract_until_next(self, needle):
         # Returns extracted bytes on success (advancing offset), or None on
         # failure
         if self._looked_for == needle:
-            search_start = max(self._start, self._looked_at - len(needle) + 1)
+            looked_at = max(self._start, self._looked_at - len(needle) + 1)
```
pgjones, line 94: As `len(len(b"\n\r?\n") == 4` (rather than 3) isn't this incorrect now? 

Although I think it should work as long as the regex representation is longer than the actual matchable length (which may not be the case)?

pgjones: Also I think we need to note that `needle` is now compiled (wasn't before) in say the docstring.

Alternatively I think this method doesn't need to take a `needle` parameter or be as generic. Maybe it is worth rewriting as `maybe_extract_until_next_line_delimiter` (or similar).

pgjones (changes requested):

h11/_receivebuffer.py
cdeler: Thank you, your explanation looks more clear (fixing)

h11/_receivebuffer.py
cdeler: > Although I think it should work as long as the regex representation is longer than the actual matchable length (which may not be the case)?

Well, yes, thinking about your comment, I found that `.*` might be much wider that 2 symbols, so I was definitely wrong in my assumption coding that

cdeler: So what I did regarding this remark:
1. I changed signature of the `maybe_extract_until_next`. Now it's
```python
def maybe_extract_until_next(self, needle_regex, max_needle_length):
```
2. it accepts pre-compiled regex and max needle length, for example:
```python
data = self.maybe_extract_until_next(blank_line_delimiter_regex, 3)
if data is None:
    return None
```
or
```python
if self._bytes_in_chunk == 0:
    # We need to refill our chunk count
    chunk_header = buf.maybe_extract_until_next(line_delimiter_regex, 2)
    if chunk_header is None:
        return None
```

cdeler:
Hello @pgjones 
I made the changes you mentioned.

Then I ran again the benchmark against the changes:

#### With changes
```
6290.6 requests/sec
6264.5 requests/sec
6302.6 requests/sec
6328.4 requests/sec
6344.2 requests/sec
6340.0 requests/sec
6347.9 requests/sec
```

#### Without changes
```
6479.7 requests/sec
6480.7 requests/sec
6502.3 requests/sec
6317.8 requests/sec
6458.3 requests/sec
6238.8 requests/sec
6278.7 requests/sec
```

h11/_receivebuffer.py
```diff
@@ -79,34 +87,57 @@ def maybe_extract_at_most(self, count):
...
+
+        return bytes(result)
+
     # HTTP/1.1 has a number of constructs where you keep reading lines until
     # you see a blank one. This does that, and then returns the lines.
     def maybe_extract_lines(self):
-        if self._data[self._start : self._start + 2] == b"\r\n":
-            self._start += 2
+        start_chunk = self._data[self._start : self._start + 2]
+        if start_chunk in [b"\r\n", b"\n"]:
```
pgjones, line 130: I think `start_chunk` can only equal `b"\n"` if `self._data == b"\n"` (given the + 2 in 129). Maybe this needs to be,
```python
if self._data[self._start : self._start + 2] == b"\r\n":
    self._start += 2
    return []
elif self._data[self._start : self._start + 1] == b"\n":
    self._start += 1
    return []
```
?

h11/_receivebuffer.py
```diff
@@ -79,34 +87,57 @@ def maybe_extract_at_most(self, count):
...
+            self._start += len(start_chunk)
             return []
         else:
-            data = self.maybe_extract_until_next(b"\r\n\r\n")
+            data = self.maybe_extract_until_next(blank_line_delimiter_regex, 3)
```
pgjones, line 134: I find it odd that `blank_line_delimiter_regex` is not `\r?\n\r?\n`, although I think the logic is find like this. If you don't think it is odd, then don't worry about it.

pgjones (changes requested):
Noticed another potential issue, I sadly missed last time.

h11/_receivebuffer.py
jab: Should it be `(\r\n\r\n|\n\n)` to be precise, on the basis that senders either always use spec-compliant line endings, or never do, rather than ever using a mix?

The `\r?\n\r?\n` version would match `"\r\n\n"` and `"\n\r\n"`.

cdeler: Well, I took `"\n\r?\n"` from this https://github.com/python-hyper/h11/issues/7#issuecomment-722640685 , but I cannot disagree that it might be a bit confusing.

I changed regex to `"(\r\n\r\n|\n\n)"` , it still should fix the original problem with a server, delimiting headers and body using double-lf

h11/_receivebuffer.py
cdeler: You are right, I missed that. Incorporated your changes

njsmith:
Looking over the data gathered in the issue history, I think the semantics we want are:

- for headers, tolerate a mix of `\r\n` and plain `\n` -- this definitely happens in practice
- I suggested searching for `\n\r?\n` because it automatically tolerates this + is simpler to match than `\r?\n\r?\n`
- for chunked-encoding, we should only support `\r\n` exactly, no need to support broken

tomchristie:
We might reduce the scope of this PR very slightly by changing the order in which we tackle things?...

* First merge https://github.com/python-hyper/h11/pull/116...
* ...and then deal with removing `.compress()` and `._start` in favour of plain `bytearray` truncation...
* ...and only then take another look at switching to this more lenient newline behaviour.

It's possible that I'm being overly cautious there, but it doesn't read to me as an easily reviewable PR just yet, and dropping out a little bit of complexity first seems sensible, right?

cdeler:
@njsmith 

Thank you for your response, I'm tuning the PR according the suggestion

The only thing which I don't understand is:
> for headers, tolerate a mix of \r\n and plain \n -- this definitely happens in practice

does it mean that we should accept headers like that
```
b"HTTP/1.1 200 OK\r\n"
b"Accept-Ranges: bytes\n"
b"Content-type: text/plain\n"
b"Connection: close\r\n"
```

For now I'm detecting the `real_lines_delimiter` using regex `b"\r?\n"` (`b"\r\n"` in the case above) and using `data.split(real_lines_delimiter)`, which is faster than `regex.split(some_bytes)`

njsmith:
> does it mean that we should accept headers like that
> 
> ```
> b"HTTP/1.1 200 OK\r\n"
> b"Accept-Ranges: bytes\n"
> b"Content-type: text/plain\n"
> b"Connection: close\r\n"
> ```

Yes, unfortunately. One of the real examples mentioned in the other thread was a server that only used `\n` for *one* header (presumably because most of the response was generated by a standard library that knows how HTTP works, and then that one header was added in by some hacky one-off code written by someone who wasn't familiar with the RFCs).

> For now I'm detecting the `real_lines_delimiter` using regex `b"\r?\n"` (`b"\r\n"` in the case above) and using `data.split(real_lines_delimiter)`, which is faster than `regex.split(some_bytes)`

Here's an idea: what if we always split on `\n`, so we get lines like:

```python
[
  b"HTTP/1.1 200 OK\r",
  b"Accept-Ranges: bytes",
  b"Content-type: text/plain",
  b"Connection: close\r",
]
```

and then we handle the trailing `\r` as part of the header name/value parsing? That already involves doing a regex match on each header line and discarding trailing whitespace...

cdeler:
Hello @njsmith ,

Sorry for the delay with an answer, looks like we have a gap between TimeZones

Wanted not to decrease `master` branch performance, I tested 3 different `maybe_extract_lines()`: one with regex split, two with an approach, suggested by @njsmith (to split lines on `b"\n"`)

### Example with `regex.split(...)`
<details>
  <summary>Example code</summary>
  
  The core part is `lines = line_delimiter_regex.split(data.rstrip(b"\r\n"))`

  ```python
    def maybe_extract_lines(self):
        if self._data[self._start : self._start + 2] == b"\r\n":
            self._start += 2
            return []
        elif self._data[self._start : self._start + 1] == b"\n":
            self._start += 1
            return []
        else:
            data = self.maybe_extract_until_next(blank_line_delimiter_regex, 3)
            if data is None:
                return None

            lines = line_delimiter_regex.split(data.rstrip(b"\r\n"))

            return lines
  ```
  
</details>

Bench result is:
```
6223.9 requests/sec
6247.8 requests/sec
6233.1 requests/sec
6250.9 requests/sec
6231.0 requests/sec
6246.4 requests/sec
6224.8 requests/sec
```

### Example with list comp `[it.rstrip(b"\r") ...]`
<details>
  <summary>Example code</summary>
  
  The core part is `lines = [it.rstrip(b"\r") for it in data.strip(b"\r\n").split(b"\n")]`

  ```python
    def maybe_extract_lines(self):
        if self._data[self._start : self._start + 2] == b"\r\n":
            self._start += 2
            return []
        elif self._data[self._start : self._start + 1] == b"\n":
            self._start += 1
            return []
        else:
            data = self.maybe_extract_until_next(blank_line_delimiter_regex, 3)
            if data is None:
                return None

            lines = [it.rstrip(b"\r") for it in data.strip(b"\r\n").split(b"\n")]

            return lines
  ```
  
</details>

Bench result is:
```
6202.6 requests/sec
6430.5 requests/sec
6444.7 requests/sec
6443.1 requests/sec
6437.4 requests/sec
6443.1 requests/sec
6415.2 requests/sec
```

### Example with `map(...)`
<details>
  <summary>Example code</summary>
  
  The core part is `lines = [it.rstrip(b"\r") for it in data.strip(b"\r\n").split(b"\n")]`

  ```python
def rstrip_line(line):
    return line.rstrip(b"\r")
...

    def maybe_extract_lines(self):
        if self._data[self._start : self._start + 2] == b"\r\n":
            self._start += 2
            return []
        elif self._data[self._start : self._start + 1] == b"\n":
            self._start += 1
            return []
        else:
            data = self.maybe_extract_until_next(blank_line_delimiter_regex, 3)
            if data is None:
                return None

            lines = list(map(rstrip_line, data.strip(b"\r\n").split(b"\n")))

            return lines
  ```
  
</details>

Bench result is:
```
6484.1 requests/sec
6497.6 requests/sec
6321.0 requests/sec
6475.9 requests/sec
6485.2 requests/sec
6526.2 requests/sec
6502.7 requests/sec
```

### Current `master` branch bench (522b00470b5776dda2dc7e98ff778eb1e418e6ad)
```
6412.4 requests/sec
6349.3 requests/sec
6545.3 requests/sec
6530.8 requests/sec
6546.8 requests/sec
6508.1 requests/sec
6421.0 requests/sec
```

## Sum up

Example with `map(...)` is the best solution I found

pgjones (approved):

cdeler:
@pgjones thank you for your approval!

I'm sorry, I hadn't seen that before I pushed the new piece of changes (aae7cdacb460cf5a14089305b84a248acc4f3d50), speeding up the bench (`map` example from  https://github.com/python-hyper/h11/pull/115#issuecomment-731624909)

h11/_receivebuffer.py
```diff
@@ -79,21 +90,29 @@ def maybe_extract_at_most(self, count):
...
         else:
-            search_start = self._start
-        offset = self._data.find(needle, search_start)
-        if offset == -1:
+            looked_at = self._start
+            self._looked_for_regex = needle_regex
+
+        delimiter_match = next(
+            self._looked_for_regex.finditer(self._data, looked_at), None
+        )
```
tomchristie, line 104: Could you help explain why this is `next(finditer(...))` rather than `search(...)` here?
What am I missing?

h11/_receivebuffer.py
```diff
@@ -79,21 +90,29 @@ def maybe_extract_at_most(self, count):
         self._start += len(out)
         return out
 
-    def maybe_extract_until_next(self, needle):
+    def maybe_extract_until_next(self, needle_regex, max_needle_length):
```
tomchristie, line 93: Should this actually be `min_needle_length`?

tomchristie: Ie. always resume searching from the first possible point where there could still be a match.

h11/_receivebuffer.py
cdeler: Thank you, you are absolutely right, I missed `Pattern::search` using `Pattern::finditer`. I'm to fix it using `search` as it's more clear

h11/_receivebuffer.py
tomchristie: Ah yeah I see it `max_needle_length`, because that pushes the restart *back* as far as possible.

tomchristie:
Sketching out @njsmith's "what if we always split on \n ... and then we handle the trailing \r as part of the header name/value parsing?", I think that'd look like this:

```python
BLANK_LINE_REGEX = re.compile(b"\n\r?\n", re.MULTILINE)


class ReceiveBuffer(object):
    def __init__(self):
        self._data = bytearray()
        self._next_line_search = 0
        self._multiple_lines_search = 0

    def __iadd__(self, byteslike):
        self._data += byteslike
        return self

    def maybe_extract_at_most(self, count):
        """
        Extract a fixed number of bytes from the buffer.
        """
        out = self._data[:count]
        if not out:
            return None

        self._data[:count] = b''
        self._next_line_search = 0
        self._multiple_lines_search = 0
        return out

    def maybe_extract_next_line(self):
        """
        Extract the first line, if it is completed in the buffer.
        """
        # Only search in buffer space that we've not already looked at.
        partial_buffer = self._data[self._next_line_search:]
        partial_idx = partial_buffer.find(b"\n")
        if partial_idx == -1:
            self._next_line_search = len(self._data)
            return None

        # Truncate the buffer and return it.
        idx = self._next_line_search + partial_idx
        out = self._data[:idx].rstrip(b"\r")
        self._data[:idx + 1] = b""
        self._next_line_search = 0
        self._multiple_lines_search = 0
        return out

    def maybe_extract_multiple_lines(self):
        """
        Extract everything up to the first blank line, and return a list of lines.
        """
        # Handle the case where we have an immediate empty line.
        if self._data[:1] == b"\n":
            self._data[:1] = b""
            self._next_line_search = 0
            self._multiple_lines_search = 0
            return []

        if self._data[:2] == b"\r\n":
            self._data[:2] = b""
            self._next_line_search = 0
            self._multiple_lines_search = 0
            return []

        # Only search in buffer space that we've not already looked at.
	partial_buffer = self._data[self._multiple_lines_search:]
        match = BLANK_LINE_REGEX.find(partial_buffer)
        if match is None:
            self._multiple_lines_search = len(self._data) - 2
            return None

        # Truncate the buffer and return it.
        idx = self._multiple_lines_search + match.span[-1]
        out = self._data[:idx]
        lines = [line.rstrip(b"\r") for line in out.split(b"\n")]
        self._data[:idx] = b""
        self._next_line_search = 0
        self._multiple_lines_search = 0
        assert lines[-2] == lines[-1] == b""
        return lines[:-2]
```

Add could potentially also drop the `rstrip(b"\n")` handling in favour of performing that within the header parsing.

I *think?* this maybe ends up reading more simply, than methods which take a general purpose regex?

I've not performance tested it our tried it out against the test cases - just a first sketch at this point.

cdeler:
@tomchristie you mean to make `maybe_extract_next_line` and `maybe_extract_multiple_lines` independent?

Well it's definitely more readable than now, e.g. do not require 2 mutual-dependent arguments (needle_regex and max regex length).

The only one thing worrying me: I'm note sure if we can truncate the buffer manually until https://github.com/python-hyper/h11/pull/116 has been merges (I'm talking about `self._data[:idx] = b""`)

tomchristie:
> I'm note sure if we can truncate the buffer manually until #116 has been merged

Possibly. Given that it's only a performance implication on Python 2, it might not necessarily need to be strictly blocked on that. We'd probably want them to occur in the same release tho.

cdeler:
@tomchristie 
Having taken your code, and slightly changed it:
1.
```diff
- self._multiple_lines_search = len(self._data) - 2
+ self._multiple_lines_search = max(0, len(self._data) - 2)
```
(without that `self._multiple_lines_search` might be negative)
2.
```diff
-        idx = self._next_line_search + partial_idx
-        out = self._data[:idx].rstrip(b"\r")
-        self._data[:idx + 1] = b""
+        idx = self._next_line_search + partial_idx + 1
+        out = self._data[:idx]
+        self._data[:idx] = b""
```
(chunk reader expects that `out` ends with `b"\r\n"`)

I estimated the performance on python3.8.5. Comparing it with the latest `master` I got the following results:
#### `master` 
```
6284.0 requests/sec
6335.5 requests/sec
6443.5 requests/sec
6456.7 requests/sec
6430.6 requests/sec
6409.8 requests/sec
6475.6 requests/sec
```

#### your code
```
6410.8 requests/sec
6406.4 requests/sec
6440.8 requests/sec
6477.5 requests/sec
6206.4 requests/sec
6397.7 requests/sec
6407.0 requests/sec
```

From my point of view it's almost equal (for both branches I've chosen the best `bench.py` run results)

tomchristie:
@cdeler Sure thing yup.

I also looked at #120 in isolation, but this PR might end up making it redundant.

tomchristie:
> chunk reader expects that out ends with b"\r\n"

That assumption might not be true here.

cdeler:
@tomchristie 
> Update: I'm referring here to the line in _readers.py, which is currently self._bytes_to_discard = 2. There's presumably no longer necessarily true after this update.

As @pgjones wrote [there](https://github.com/python-hyper/h11/pull/115#issuecomment-731139500)

> for chunked-encoding, we should only support \r\n exactly, no need to support broken peers that use \n

I think in this case `maybe_extract_next_line` should search for `b"\r\n"`. Then the function might look like:
```python
    def maybe_extract_next_line(self):
        """
        Extract the first line, if it is completed in the buffer.
        """
        # Only search in buffer space that we've not already looked at.
        search_start_index = max(0, self._next_line_search - 1)
        partial_buffer = self._data[search_start_index :]
        partial_idx = partial_buffer.find(b"\r\n")
        if partial_idx == -1:
            self._next_line_search = len(self._data)
            return None

        # Truncate the buffer and return it.
        # 2 == len(b"\r\n")
        idx = search_start_index + partial_idx + 2
        out = self._data[:idx]
        self._data[:idx] = b""
        self._next_line_search = 0
        self._multiple_lines_search = 0
        return out
```

cdeler:
@tomchristie @njsmith 

Sorry for disturbing you, but is there any estimated date regarding this PR & next release processing flow? May be I can do something to make the reviewing easier?

I'm asking since some users from one downstream project (`home-assistant`)  are suffering with problems, when some sensors or devices are not accessible due to some problems with line delimiters (e.g. https://github.com/home-assistant/core/issues/42415)

tomchristie:
First up...

Based on this: https://github.com/home-assistant/core/issues/42415#issuecomment-731127651

Have we *confirmed* that the issue they're seeing is resolved by handling non-RFC-compliant line endings?

If so I can certainly make some more review time available, and help get this towards a release.

cdeler:
@tomchristie 
The credentials provided by @Kane610 produces 401, so I still haven't verified the latest fix with an original device, but I can confirm that I checked the original PR versus one affected device, and the changeset fixed the problem. 

I'll try to ping @Kane610 or someone in the https://github.com/home-assistant/core/issues/42415 to obtain credentials and check the latest changes against the real device

cdeler:
Hello @tomchristie

Given the credentials provided by @wonderiuy, I checked the fix from this PR.

I can confirm that this fix solves an error with access to some `axis` resource

### How I checked?
#### `requests` lib (they used it before):
```
In [1]: import requests
In [2]: from requests.auth import HTTPDigestAuth
In [3]: url = "http://[REDACTED]/axis-cgi/pwdgrp.cgi?action=get"
In [4]: requests.get(url, auth=HTTPDigestAuth("[REDACTED]", "[REDACTED]"))
Out[4]: <Response [200]>
```

#### `httpx` without fix:
```
In [1]: import httpx
In [2]: url = "http://[REDACTED]/axis-cgi/pwdgrp.cgi?action=get"
In [3]: auth = httpx.DigestAuth("[REDACTED]", "[REDACTED]")
In [4]: client = httpx.Client(auth=auth)
In [5]: client.get(url)
RemoteProtocolError: peer unexpectedly closed connection
```

#### `httpx` with fix:
```
In [1]: import httpx
In [2]: url = "http://[REDACTED]/axis-cgi/pwdgrp.cgi?action=get"
In [3]: auth = httpx.DigestAuth("[REDACTED]", "[REDACTED]")
In [4]: client = httpx.Client(auth=auth)
In [5]: client.get(url)
Out[5]: <Response [200 OK]>
```

Kane610:
How would I go about if I'd want to test cdelers fix locally? I'd be happy to verify as well.

wonderiuy:
You can use the address and credentials i've given to you, firmware on that device is 5.20.5

Kane610:
@wonderiuy I meant getting his changes to my local filesystem :)

Kane610:
It was really simple with GitHub desktop to clone this branch and try it out. Works well on my 5.51 device at home. Good job @cdeler !

donnib:
I can also confirm this is working locally ;) Looking forward for the PR to be approved.

hoorna:
Last week Axis published a new firmware version (5.51.7.2) for my camera (M1034-W). Below you will find the release notes.

Don't know if it has something to do with the issue (LF vs CRLF) in this thread but the first correction (C01)  in the new firmware version is handling about "Corrected a newline character". 

I thought it would be wise to make a notice about it in this tread.

------------------------------------------------------------------------------------------

Corrections in 5.51.7.2 since 5.51.7.1
=======================================

5.51.7.2:C01
Corrected a newline character in pwdgrp.cgi, introduced in 5.51.6, that could cause
problems when parsing the response.

5.51.7.2:C02
Corrected an issue that prevented Action Rule Events from sending images via email.

5.51.7.2:C03
Corrected an issue that caused monolith to timeout and respawn during too many
connect/disconnect RTSP streaming requests.

5.51.7.2:C04
Added support to enable/disable X-Frame-Options headers in the plainconfig. By default,
X-Frame-Options is enabled and its value is set to "sameorigin".

Kane610:
> Last week Axis published a new firmware version (5.51.7.2) for my camera (M1034-W). Below you will find the release notes.
> 
> Don't know if it has something to do with the issue (LF vs CRLF) in this thread but the first correction (C01) in the new firmware version is handling about "Corrected a newline character".

Thanks! I don't know, but regardless there are other firmwares that don't get this update so it is still needed.

wonderiuy:
I've checked with the new firmware 5.51.7.2 but the problem is still there, so the fix from @cdeler and @Kane610 is more than welcome :)

astrandb:
Can also confirm that this PR solves the problems with older versions of Axis cameras in Homeassistant.

Kane610:
Hey guys! Any progress on getting this merged?

h11/_receivebuffer.py
```diff
@@ -38,75 +40,101 @@
...
-            if data is None:
-                return None
-            lines = data.split(b"\r\n")
-            assert lines[-2] == lines[-1] == b""
-            del lines[-2:]
-            return lines
+
+        # Only search in buffer space that we've not already looked at.
+        partial_buffer = self._data[self._multiple_lines_search :]
+        match = blank_line_regex.search(partial_buffer)
```
njsmith, line 124: Instead of copying the buffer, use [the `pos` argument to search](https://docs.python.org/3/library/re.html#re.Pattern.search):

```python
match = blank_line_regex.search(self._data, self._multiple_lines_search)
```

h11/_receivebuffer.py
```diff
@@ -38,75 +40,101 @@
...
+    def maybe_extract_next_line(self):
+        """
+        Extract the first line, if it is completed in the buffer.
+        """
+        # Only search in buffer space that we've not already looked at.
+        search_start_index = max(0, self._next_line_search - 1)
+        partial_buffer = self._data[search_start_index:]
+        partial_idx = partial_buffer.find(b"\r\n")
+        if partial_idx == -1:
+            self._next_line_search = len(self._data)
```
njsmith, line 93: In `maybe_extract_next_line`, we store the raw buffer length in `self._next_line_search` and then do the subtraction when we use it. In `maybe_extract_lines`, we store the "pre-subtracted" value, so we can use it directly. This inconsistency is kind of confusing :-). We should switch one of them so they match.

h11/_receivebuffer.py
```diff
@@ -38,75 +40,101 @@
...
+        Extract a fixed number of bytes from the buffer.
+        """
+        out = self._data[:count]
         if not out:
             return None
-        self._start += len(out)
+
+        self._data[:count] = b""
+        self._next_line_search = 0
+        self._multiple_lines_search = 0
```
njsmith, line 81: There's a lot of copy/pastes of this code for extracting an initial slice and then doing internal bookkeeping, which is error prone. Let's factor it out into a method, like:

```python
def _extract(self, count):
    out = self._data[:count]
    del self._data[:count]
    self._next_line_search = 0
    self._multiple_lines_search = 0
```

And then in all the other methods, just do:

```python
    return self._extract(whatever_length_value_we_ended_up_with)
```

h11/tests/test_receivebuffer.py
```diff
@@ -76,3 +81,54 @@ def test_receivebuffer():
...
+        b += line
+
+    lines = b.maybe_extract_lines()
+
+    assert lines == [
+        b"HTTP/1.1 200 OK",
+        b"Content-type: text/plain",
+        b"Connection: close",
+    ]
+    assert bytes(b) == b"Some body"
```
njsmith, line 134: Can you also add a few similar tests to `test_readers_unusual` in `test_io.py`? These tests are good for the basic `ReceiveBuffer` functionality, but the test harness there sets up a more "end-to-end" setup that runs our full http parsing pipeline, so it would give us more confidence that everything is wired up correctly.

h11/_receivebuffer.py
```diff
@@ -38,75 +40,101 @@
...
+        partial_buffer = self._data[self._multiple_lines_search :]
+        match = blank_line_regex.search(partial_buffer)
+        if match is None:
+            self._multiple_lines_search = max(0, len(self._data) - 2)
+            return None
+
+        # Truncate the buffer and return it.
+        idx = self._multiple_lines_search + match.span(0)[-1]
+        out = self._data[:idx]
+        lines = [line.rstrip(b"\r") for line in out.split(b"\n")]
```
njsmith, line 132: Instead of calling `rstrip` here (which always has to copy the whole buffer, since these are bytearrays), I think we could leave the trailing `\r` in, and then in `_abnf.py` change `header_field`, `request_line`, and `status_line` to match a trailing optional `\r`, e.g.:

```python
header_field = (
    r"(?P<field_name>{field_name})"
    r":"
    r"{OWS}"
    r"(?P<field_value>{field_value})"
    r"{OWS}\r?".format(**globals())         # <-- notice added \r? here
)
```

njsmith:
Oh also, forgot to say: could you also add a news entry for the next release notes? See `newsfragments/README.rst` for the details on how to do that.

h11/_receivebuffer.py
cdeler: I introduced the `_extract`, and replaced all copy/pastes using this function.
It also helped to resolve the above comment (https://github.com/python-hyper/h11/pull/115#discussion_r540085272)

h11/_receivebuffer.py
cdeler: Hello @njsmith ,

I tried to do that.

I added `\r?` to these regexes (`header_field`, and `request_line`, and `status_line`)

Then I set
```python
lines = out.split(b"\n")
```

but it broke one of `test_readers_unusual` test cases in `test_io.py`
<details>
  <summary>the test source</summary>

```python
def test_readers_unusual():
    ...

    # obsolete line folding
    tr(
        READERS[CLIENT, IDLE],
        b"HEAD /foo HTTP/1.1\r\n"
        b"Host: example.com\r\n"
        b"Some: multi-line\r\n"
        b" header\r\n"
        b"\tnonsense\r\n"
        b"    \t   \t\tI guess\r\n"
        b"Connection: close\r\n"
        b"More-nonsense: in the\r\n"
        b"    last header  \r\n\r\n",
        Request(
            method="HEAD",
            target="/foo",
            headers=[
                ("Host", "example.com"),
                ("Some", "multi-line header nonsense I guess"),
                ("Connection", "close"),
                ("More-nonsense", "in the last header"),
            ],
        ),
    )
``` 

</details>

The header
```python
        b"Some: multi-line\r\n"
        b" header\r\n"
        b"\tnonsense\r\n"
        b"    \t   \t\tI guess\r\n"
```
turns to
```python
        b"Some: multi-line\r header\r\tnonsense\r    \t   \t\tI guess\r\n"
```

I cannot figure out how to carefully cut out the `\r` from such a line.

cdeler:
Hello @njsmith ,

I resolved all remarks apart from [this](https://github.com/python-hyper/h11/pull/115#discussion_r540096182).

I cannot understand how to clearly and carefully fix the broken test case (probably we cannot just remove it).

h11/_receivebuffer.py
cdeler: As soon as I rewrote both methods using `_extract(...)` method, this issue resolved (both methods works with offsets)

h11/_receivebuffer.py
cdeler: After some discussion in [gitter](https://gitter.im/python-hyper/community?at=5fd7843082222960e4e589fd), it has been decided not to change regexes, but to rewrite `.rstrip(...)` with 
```python
        for line in lines:
            if line.endswith(b"\r"):
                del line[-1]
```
to avoid extra memory allocation

njsmith:
@cdeler I tweaked your newsfragment to use the correct quoting: in ReST, code literals require *double* backticks. Super annoying if you're used to markdown, but what can ya do.

@pgjones Note that this PR doesn't quite drop support for py2, but it does change the buffer handling to be O(n**2) on py2, and I'm wondering if we should flag that in the release notes or anything. Or are you planning to drop py2 for real in the next release anyway?

pgjones:
Lets merge this, and drop Py2 for the next release.

wonderiuy:
This is like a Christmas gift, a big thank you to every1 involved

PR #116, opened 2020-11-18
Drop support for Python 2

### LINKED ISSUE #114: Drop support for Python 2?
I'd like to drop support for Python 2, type hint this codebase, and make use of features like dataclasses. I've a local version where I've done this - to allow testing of mypyc (it does speed things up). I know around this time last year there was a need/desire to keep Python 2 support for urllib3 (have I remembered this correctly?). Is there still a motivating need for Python 2 support? 

To clarify I think the status quo is currently fine, however if Python 2 isn't needed/desired then great.

bluetech:
Blocked on #115 / 0.12.0 release, will be updated afterwards.

Fixes #114.

Python 2 support in this library is not too onerous, the main reason for doing this is to allow using newer Python features like inline type annotations.

In the documentation/README updates I assume that this will be released in a 0.12.x series rather than the latest 0.11.x series which supports Python 2. I think it would be good to branch off 0.11.x just in case we ever want to backport some bug fixes to a Python 2-compatible version.

I updated whatever I could find except for the `ReceiveBuffer. According to the [comment](https://github.com/python-hyper/h11/blob/522b00470b5776dda2dc7e98ff778eb1e418e6ad/h11/_receivebuffer.py#L20-L40) there it can be simplified after Python 2 is dropped but needs some care. I will look at it in a separate PR unless someone else wants to update this part.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -112,7 +112,8 @@ library.
 It has a test suite with 100.0% coverage for both statements and
 branches.
 
-Currently it supports Python 3 (testing on 3.5-3.8), Python 2.7, and PyPy.
+Currently it supports Python 3 (testing on 3.5-3.8) and PyPy 3.
+The last Python 2-compatible version was h11 0.11.x.
 (Originally it had a Cython wrapper for `http-parser
 <https://github.com/nodejs/http-parser>`_ and a beautiful nested state
 machine implemented with ``yield from`` to postprocess the output. But
--- a/docs/source/index.rst
+++ b/docs/source/index.rst
@@ -44,7 +44,9 @@ whatever. But h11 makes it much easier to implement something like
 Vital statistics
 ----------------
 
-* Requirements: Python 2.7 or Python 3.5+ (PyPy works great)
+* Requirements: Python 3.5+ (PyPy works great)
+
+  The last Python 2-compatible version was h11 0.11.x.
 
 * Install: ``pip install h11``
 
--- a/newsfragments/114.removal.rst
+++ b/newsfragments/114.removal.rst
@@ -1,0 +1,2 @@
+Python 2.7 and PyPy 2 support is removed. h11 now requires Python>=3.5 including PyPy 3.
+Users running `pip install h11` on Python 2 will automatically get the last Python 2-compatible version.
--- a/setup.py
+++ b/setup.py
@@ -17,15 +17,15 @@ setup(
     # This means, just install *everything* you see under h11/, even if it
     # doesn't look like a source file, so long as it appears in MANIFEST.in:
     include_package_data=True,
+    python_requires=">=3.5",
     classifiers=[
         "Development Status :: 3 - Alpha",
         "Intended Audience :: Developers",
         "License :: OSI Approved :: MIT License",
         "Programming Language :: Python :: Implementation :: CPython",
         "Programming Language :: Python :: Implementation :: PyPy",
-        "Programming Language :: Python :: 2",
-        "Programming Language :: Python :: 2.7",
         "Programming Language :: Python :: 3",
+        "Programming Language :: Python :: 3 :: Only",
         "Programming Language :: Python :: 3.5",
         "Programming Language :: Python :: 3.6",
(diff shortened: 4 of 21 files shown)


pquentin (approved):

bluetech:
(I added a change to remove an `__ne__` implementation - since py3 it's auto implemented based on `__eq__`)

README.rst
```diff
@@ -112,7 +112,8 @@ library.
 It has a test suite with 100.0% coverage for both statements and
 branches.
 
-Currently it supports Python 3 (testing on 3.5-3.8), Python 2.7, and PyPy.
+Currently it supports Python 3 (testing on 3.5-3.8) and PyPy 3.
+The last Python 2-compatible version was h11 0.11.x.
```
pgjones, line 116: 0.12.0, as I don't think the line change behavior counts as a bug fix.

docs/source/index.rst
```diff
@@ -44,7 +44,9 @@ whatever. But h11 makes it much easier to implement something like
 Vital statistics
 ----------------
 
-* Requirements: Python 2.7 or Python 3.5+ (PyPy works great)
+* Requirements: Python 3.5+ (PyPy works great)
```
pgjones, line 47: 3.6+

docs/source/index.rst
```diff
@@ -44,7 +44,9 @@ whatever. But h11 makes it much easier to implement something like
...
+
+  The last Python 2-compatible version was h11 0.11.x.
```
pgjones, line 49: 0.12.0 (to remember).

newsfragments/114.removal.rst
```diff
@@ -0,0 +1,2 @@
+Python 2.7 and PyPy 2 support is removed. h11 now requires Python>=3.5 including PyPy 3.
```
pgjones, line 1: 3.6+

setup.py
```diff
@@ -17,15 +17,15 @@
     # This means, just install *everything* you see under h11/, even if it
     # doesn't look like a source file, so long as it appears in MANIFEST.in:
     include_package_data=True,
+    python_requires=">=3.5",
```
pgjones, line 20: 3.6

setup.py
```diff
@@ -17,15 +17,15 @@
...
         "Development Status :: 3 - Alpha",
         "Intended Audience :: Developers",
         "License :: OSI Approved :: MIT License",
         "Programming Language :: Python :: Implementation :: CPython",
         "Programming Language :: Python :: Implementation :: PyPy",
-        "Programming Language :: Python :: 2",
-        "Programming Language :: Python :: 2.7",
         "Programming Language :: Python :: 3",
+        "Programming Language :: Python :: 3 :: Only",
         "Programming Language :: Python :: 3.5",
```
pgjones, line 29: Can you also remove this?

pgjones (approved):
Some minor versioning comments. I think we state 3.6+ given 3.5 is EOL.

Otherwise this is great, thanks!

bluetech:
@pgjones

> I think we state 3.6+ given 3.5 is EOL.

Sounds good to me, would definitely be a little nicer for typing. Should I open an issue for discussion or just add it here? (Maybe some dependents still want to support Python 3.5)

> 0.12.0, as I don't think the line change behavior counts as a bug fix.

I'll mark this PR as blocked on #115/0.12.0, will update after they're done.

pquentin:
(It would be nice to test 3.9 in CI as well)

tomchristie:
> I'll mark this PR as blocked on #115/0.12.0, will update after they're done.

I reckon we should maybe get these done the other way around, no?

A progression that looks sensible to me is:

* Pull in this first.
* Then get #120 merged, because at that point it's a straight-forward performance improvement.
* Then look at #116, because it'll have a slightly smaller change footprint.

pgjones:
Yea, I think that is the better flow.

README.rst
pgjones: This is now correct.

pgjones:
Merged via command line (fixing conflicts).

PR #118, opened 2020-11-18
Get rid of _EventBundle

bluetech:
This is based on #116 so you can ignore the first commit.

`_EventBundle` uses a lot of dynamic python features to save on some duplication, but it slows things down, and will also make it much harder to add static typing. Since these types are now pretty stable, it seems not worth it.

On the bench/ micro-benchmark:

```
Before:  9322.6 requests/sec
After : 10544.6 requests/sec
```

### DIFF
--- a/h11/_events.py
+++ b/h11/_events.py
@@ -24,7 +24,7 @@ __all__ = [
 request_target_re = re.compile(request_target.encode("ascii"))
 
 
-class _EventBundle(object):
+class _EventBundle:
     _fields = []
     _defaults = {}
 
@@ -84,9 +84,6 @@ class _EventBundle(object):
     # Useful for tests
     def __eq__(self, other):
         return self.__class__ == other.__class__ and self.__dict__ == other.__dict__
-
-    def __ne__(self, other):
-        return not self.__eq__(other)
 
     # This is an unhashable type.
     __hash__ = None
--- a/h11/_events.py
+++ b/h11/_events.py
@@ -24,72 +24,7 @@ __all__ = [
 request_target_re = re.compile(request_target.encode("ascii"))
 
 
-class _EventBundle:
-    _fields = []
-    _defaults = {}
-
-    def __init__(self, **kwargs):
-        _parsed = kwargs.pop("_parsed", False)
-        allowed = set(self._fields)
-        for kwarg in kwargs:
-            if kwarg not in allowed:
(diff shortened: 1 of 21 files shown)


bluetech:
@pgjones if this conflicts with something you're doing let me know! (In #114 you mentioned converting to dataclasses, but these require py37 and I personally prefer to avoid them in public library API).

pgjones:
I'm happy with this, I've no strong attachment to dataclasses. I'm minded to merge after #116.

h11/_state.py
```diff
@@ -197,7 +197,7 @@
 }
 
 
-class ConnectionState(object):
+class ConnectionState:
```
sethmlarson, line 200: Should we use `__slots__` for this object as well since it's accessed frequently?

sethmlarson (commented):
Seems good to me, one question:

h11/_state.py
bluetech: I'm in favor of adding slots to everything personally. I'll do it in a separate PR after some of the others are settled.

h11/_events.py
```diff
@@ -146,13 +87,31 @@ def _validate(self):
...
+            self.headers,
+            self.http_version,
+        )
 
-class _ResponseBase(_EventBundle):
-    _fields = ["status_code", "headers", "http_version", "reason"]
-    _defaults = {"http_version": b"1.1", "reason": b""}
+    # Useful for tests
+    def __eq__(self, other):
+        if not isinstance(other, type(self)):
```
pgjones, line 101: Whilst this changes the logic of what is equal or not (from `self.__class__ == other.__class__`) to include subclasses and potentially raising if incompatible classes are compared (rather than returning False) I think it is fine as this is mostly (and likely only) used for tests.

pgjones:
Having said I've no attachment to dataclasses; what do you think of the usage of `frozen` as in [wsproto](https://github.com/python-hyper/wsproto/blob/master/src/wsproto/events.py#L23)? I think that better clarifies the API for these objects.

sethmlarson:
I'm +1 on using dataclasses and frozen.

pgjones:
See #124 as the dataclass alternative.

pgjones:
I've merged the dataclass approach for the reasons discussed, plus that we have a desire to use dataclasses consistently in the hyper projects.

PR #119, opened 2020-11-18
Inline the validate() function

bluetech:
This is based on #118, please look at the last commit only. (Sorry for the jungle of PRs, I can rebase them whichever way).

---

It somewhat obscures the control flow and adds some non-trivial overhead.

Inline it in favor of the direct match/if combination.

On the bench/ microbenchmark:

```
Before: 10700.2 requests/sec
After : 11334.0 requests/sec
```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,11 +13,9 @@ jobs:
       max-parallel: 5
       matrix:
         python-version:
-        - 2.7
         - 3.6
         - 3.7
         - 3.8
-        - pypy2
         - pypy3
 
     steps:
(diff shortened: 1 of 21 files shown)


pgjones:
I also prefer this approach, however as it more of a preference question I'll wait for more views.

tomchristie:
> I also prefer this approach, however as it more of a preference question I'll wait for more views.

From the last commit, that seems *okay* as a trade-off, yup.
Tho I'd probably also want to review again once there's an up to date PR for it, and have the micorbenchmark re-run against the latest PR-against-master.

PR #120, opened 2020-11-24
Use native bytearray truncation

tomchristie:
Switching from `.compress()` to using native buffer truncation.

I figure despite some related work on #115 it's worth looking at this PR in isolation.

We probably want this to be blocked on #116.

Benchmarking on Python 3.7...

Before:

```python
$ PYTHONPATH=. venv/bin/python bench/benchmarks/benchmarks.py
6901.9 requests/sec
7055.0 requests/sec
7084.2 requests/sec
7070.6 requests/sec
7107.9 requests/sec
7075.8 requests/sec
7079.4 requests/sec

$ PYTHONPATH=. venv/bin/python bench/benchmarks/benchmarks.py
6972.9 requests/sec
6996.9 requests/sec
6947.3 requests/sec
6999.9 requests/sec
7040.6 requests/sec
7030.0 requests/sec
6997.6 requests/sec
```

After:

```python
$ PYTHONPATH=. venv/bin/python bench/benchmarks/benchmarks.py
7054.6 requests/sec
7133.3 requests/sec
7148.3 requests/sec
7116.4 requests/sec
7121.0 requests/sec
7128.2 requests/sec
7156.5 requests/sec

$ PYTHONPATH=. venv/bin/python bench/benchmarks/benchmarks.py
7018.9 requests/sec
7074.7 requests/sec
7075.0 requests/sec
7070.6 requests/sec
7028.2 requests/sec
7118.3 requests/sec
7131.2 requests/sec
```

### DIFF
--- a/h11/_receivebuffer.py
+++ b/h11/_receivebuffer.py
@@ -45,8 +45,8 @@ class ReceiveBuffer(object):
         out = self._data[:count]
         if not out:
             return None
-        # Note that front-truncation of bytesarray is O(1), from Python 3.4
-        # onwards, thanks to some excellent work by Antoine Pitrou:
+        # Note that front-truncation of bytesarray is amortized O(1), from
+        # Python 3.4 onwards, thanks to some excellent work by Antoine Pitrou:
         #
         # https://bugs.python.org/issue19087
         del self._data[:count]
--- a/h11/_receivebuffer.py
+++ b/h11/_receivebuffer.py
@@ -45,7 +45,11 @@ class ReceiveBuffer(object):
         out = self._data[:count]
         if not out:
             return None
-        self._data[:count] = b""
+        # Note that front-truncation of bytesarray is amortized O(n), from
+        # Python 3.4 onwards, thanks to some excellent work by Antoine Martin:
+        #
+        # https://bugs.python.org/issue19087
+        del self._data[:count]
         return out
 
     def maybe_extract_until_next(self, needle):
@@ -62,14 +66,14 @@ class ReceiveBuffer(object):
             return None
         new_start = offset + len(needle)
         out = self._data[:new_start]
-        self._data[:new_start] = b""
+        del self._data[:new_start]
         return out
 
     # HTTP/1.1 has a number of constructs where you keep reading lines until
     # you see a blank one. This does that, and then returns the lines.
     def maybe_extract_lines(self):
         if self._data[:2] == b"\r\n":
-            self._data[:2] = b""
+            del self._data[:2]
             return []
         else:
             data = self.maybe_extract_until_next(b"\r\n\r\n")
--- a/h11/_receivebuffer.py
+++ b/h11/_receivebuffer.py
@@ -42,7 +42,6 @@ class ReceiveBuffer(object):
     def __init__(self):
         self._data = bytearray()
         # These are both absolute offsets into self._data:
-        self._start = 0
         self._looked_at = 0
         self._looked_for = b""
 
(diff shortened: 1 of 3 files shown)


tomchristie:
We also ought to follow up on the action in the comment "make sure PyPy also has the optimization", but I'm not sure where to start with that.

h11/_receivebuffer.py
```diff
@@ -16,33 +16,10 @@
 #   of constantly copying
 # WARNING:
 # - I haven't benchmarked or profiled any of this yet.
-#
-# Note that starting in Python 3.4, deleting the initial n bytes from a
-# bytearray is amortized O(n), thanks to some excellent work by Antoine
-# Martin:
-#
-#     https://bugs.python.org/issue19087
```
bluetech, line 24: I think it'd be helpful to keep the paragraph above.

h11/_receivebuffer.py
```diff
@@ -51,56 +28,48 @@ def __bool__(self):
...
         self._data += byteslike
         return self
 
     def maybe_extract_at_most(self, count):
-        out = self._data[self._start : self._start + count]
+        out = self._data[:count]
         if not out:
             return None
-        self._start += len(out)
+        self._data[:count] = b""
```
bluetech, line 48: I think a `del` would be a little clearer but maybe it's just me. (Also below).

tomchristie: Agreed, yup!

h11/_receivebuffer.py
tomchristie: Fair point, yes. I've addressed it as an inline comment.

h11/_receivebuffer.py
```diff
@@ -51,56 +28,52 @@ def __bool__(self):
...
         self._data += byteslike
         return self
 
     def maybe_extract_at_most(self, count):
-        out = self._data[self._start : self._start + count]
+        out = self._data[:count]
         if not out:
             return None
-        self._start += len(out)
+        # Note that front-truncation of bytesarray is amortized O(n), from
```
bluetech, line 48: The comment is a bit confusing I guess; the truncation itself is amortized O(1), I think what the comment wanted to say is that it makes the entire process O(n), but I would change here to O(1).

bluetech:
>We also ought to follow up on the action in the comment "make sure PyPy also has the optimization", but I'm not sure where to start with that.

With this code:

```py
size = 50 * 2**20
b = bytearray(size)
for i in range(size):
    del b[0]
```

Results:

```
python3.8:           4.257s
pypy3 version 7.3.2: 0.326s
pypy2 version 7.3.2: 0.326s
python2.7:           DNF
```

So it's either that pypy has this optimization, or it's doing some magic JIT handling of this specific code. I tried some tricks to foil any such optimization and it seems legit.

If we're willing to have a timing-based unit test for this, a `size = 3 * 2**20` with a timeout of 5s would do the trick IMO, even with a super fast CPU running the quadratic case or a very slow CPU running the linear case. But it will probably be flaky anyway...

h11/_receivebuffer.py
```diff
@@ -51,56 +28,52 @@ def __bool__(self):
...
         return self
 
     def maybe_extract_at_most(self, count):
-        out = self._data[self._start : self._start + count]
+        out = self._data[:count]
         if not out:
             return None
-        self._start += len(out)
+        # Note that front-truncation of bytesarray is O(1), from Python 3.4
+        # onwards, thanks to some excellent work by Antoine Pitrou:
```
bluetech, line 49: I suggest keeping it as "amortized O(1)" , a bit more instructive and technically correct:

```suggestion
        # Note that front-truncation of bytesarray is amortized O(1), from
        # Python 3.4 onwards, thanks to some excellent work by Antoine Pitrou:
```

bluetech (approved):

njsmith:
I think this has been superseded by #115

pgjones:
Agree with #115 superseding this. (Please reopen if not).

PR #122, opened 2020-12-20
Add early detection of invalid http data when request line starts with binary

### LINKED ISSUE #121: Server wait forever on obviously incorrect but incomplete HTTP request
I've encountered this behavior by initiating an HTTPS request with a client to a server expecting HTTP.
So the server end up with a TLS "Client Hello" frame which doesn't contains the `\r\n\r\n` pattern that should mark the end of the header part of the HTTP request.

Hence h11 just consider the request is not over yet, hence the server starts waiting forever while the client is waiting for a TLS "Server Hello" answer :'(

https://github.com/python-hyper/h11/blob/522b00470b5776dda2dc7e98ff778eb1e418e6ad/h11/_readers.py#L70-L81

https://github.com/python-hyper/h11/blob/522b00470b5776dda2dc7e98ff778eb1e418e6ad/h11/_receivebuffer.py#L101-L112

A quick&dirty test on various websites (heroku.com using cowboy, yahoo.com using ATS and google.com) shows that parsing of the HTTP request start line occurs as soon as it is received (and so a `400 Invalid HTTP Request` is returned instead of waiting indefinitely).

I didn't check if this behavior is specified in the RFC (I guess it's not ^^), however it seems a legit behavior... or at least something worth talking about ;-)

touilleMan:
closes #121

### DIFF
--- a/h11/_readers.py
+++ b/h11/_readers.py
@@ -64,6 +64,8 @@ request_line_re = re.compile(request_line.encode("ascii"))
 def maybe_read_from_IDLE_client(buf):
     lines = buf.maybe_extract_lines()
     if lines is None:
+        if buf.is_next_line_obviously_invalid_request_line():
+            raise LocalProtocolError("illegal request line")
         return None
     if not lines:
         raise LocalProtocolError("no request line received")
@@ -81,6 +83,8 @@ status_line_re = re.compile(status_line.encode("ascii"))
 def maybe_read_from_SEND_RESPONSE_server(buf):
     lines = buf.maybe_extract_lines()
     if lines is None:
+        if buf.is_next_line_obviously_invalid_request_line():
+            raise LocalProtocolError("illegal request line")
         return None
     if not lines:
         raise LocalProtocolError("no response line received")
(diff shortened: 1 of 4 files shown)


njsmith:
Looks good to me, just two nitpicks:

- Can you add a newsfragment?
- Can you tweak the comment (and maybe the test) to note that client TLS handshakes always start with 0x16, so this is 100% guaranteed to catch them? (It's not even a protocol version, it's a "message type" code – 0x16 is the code for "this is a handshake message", and it's been consistent across all versions of TLS so far.)

sethmlarson:
Since the primary case we're covering is TLS should we test 0x16 explicitly? Otherwise LGTM.

touilleMan:
@njsmith @sethmlarson I've updated my PR ;-)

pgjones:
Could you update again (note there is an actual conflict here, self._start is no longer used)?

touilleMan:
@pgjones rebased ! \o/

(and happy new year to the h11 team :partying_face: :champagne:  :tada: )

PR #124, opened 2020-12-27
Switch event classes to dataclasses

pgjones:
The _EventBundle used previously dynamically creates the Event
classes, which makes typing hard. As the Event classes are now stable
the dynamic creation isn't required, and as h11 supports Python3.6+
dataclasses can be used instead.

This now makes the Events frozen (and almost imutable) which better
matches the intended API. In turn it requires the `object.__setattr__`
usage and the alteration to the
`_clean_up_response_headers_for_sending` method.

This change also improves the performance a little, using the
benchmark,

Before: 6.7k requests/sec
After: 6.9k requests/sec

Notes:

The test response-header changes are required as the previous version
would mutate the response object.

The init for the Data event is required as slots and defaults aren't
possible with dataclasses.

### DIFF
--- a/h11/_events.py
+++ b/h11/_events.py
@@ -6,13 +6,17 @@
 # Don't subclass these. Stuff will break.
 
 import re
+from abc import ABC
+from dataclasses import dataclass, field
+from typing import Any, cast, Dict, List, Tuple, Union
 
-from . import _headers
 from ._abnf import request_target
+from ._headers import Headers, normalize_and_validate
 from ._util import bytesify, LocalProtocolError, validate
 
 # Everything in __all__ gets re-exported as part of the h11 public API.
 __all__ = [
+    "Event",
     "Request",
     "InformationalResponse",
     "Response",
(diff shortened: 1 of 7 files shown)


h11/_events.py
```diff
@@ -24,72 +28,16 @@
...
-        kwarg_str = ", ".join(kwarg_strs)
-        return "{}({})".format(name, kwarg_str)
-
-    # Useful for tests
-    def __eq__(self, other):
-        return self.__class__ == other.__class__ and self.__dict__ == other.__dict__
+class Event(ABC):
+    """
+    Base class for h11 events.
+    """
```
bluetech, line 35: In order for `__slots__` to eliminate the `__dict__` it needs to be added in the base class as well.

bluetech (commented):
I am personally not a big fun of using dataclasses or attrs in a public library API, because

1. They add some startup overhead (probably negligible in this case though).
2. They add "hidden" APIs to the classes, like `dataclasses.asdict`, `dataclasses.replace` etc. which users will eventually rely on, making it part of the public API.
3. They somewhat reduce (or add friction to) the library's control should it ever need it.
4. They tend to do some funky stuff, for example in pytest we had some trouble with what dataclasses does [here](https://github.com/python/cpython/blob/897387d2c8a956e74770c6bdd0447dfec61e8e27/Lib/dataclasses.py#L615).

That said, if you're OK with this, the code LGTM except one comment.

h11/_events.py
pgjones: Thanks, fixed this.

PR #125, opened 2020-12-27
Clarify that the Headers class is a Sequence

pgjones:
This is both in the docs (via the changelog) and by inheriting from
the Sequence ABC.


bluetech (approved):
LGTM.

Optionally, `Sequence` has an `__iter__` mixin so you could remove the manual implementation if you want.

PR #135, opened 2021-08-28
Add typing and enforce checking via tox/CI

pgjones:
This uses the same mypy settings as wsproto.

The Sentinel values are problematic, but I've found no good solution
that also has the property that type(Sentinel) is Sentinel - so this
should suffice for now.

Whilst I've been lazy with the tests, I've mostly avoided type ignores
in the main code. This should ensure that mypyc can be used if
desired.

### DIFF
--- a/h11/__init__.py
+++ b/h11/__init__.py
@@ -6,16 +6,57 @@
 # semantics to check that what you're asking to write to the wire is sensible,
 # but at least it gets you out of dealing with the wire itself.
 
-from ._connection import *
-from ._events import *
-from ._state import *
-from ._util import LocalProtocolError, ProtocolError, RemoteProtocolError
-from ._version import __version__
+from h11._connection import Connection, NEED_DATA, PAUSED
+from h11._events import (
+    ConnectionClosed,
+    Data,
+    EndOfMessage,
+    Event,
+    InformationalResponse,
+    Request,
+    Response,
+)
+from h11._state import (
+    CLIENT,
+    CLOSED,
+    DONE,
+    ERROR,
+    IDLE,
+    MIGHT_SWITCH_PROTOCOL,
+    MUST_CLOSE,
+    SEND_BODY,
+    SEND_RESPONSE,
+    SERVER,
+    SWITCHED_PROTOCOL,
+)
+from h11._util import LocalProtocolError, ProtocolError, RemoteProtocolError
+from h11._version import __version__
 
 PRODUCT_ID = "python-h11/" + __version__
 
 
-__all__ = ["ProtocolError", "LocalProtocolError", "RemoteProtocolError"]
-__all__ += _events.__all__
-__all__ += _connection.__all__
-__all__ += _state.__all__
+__all__ = (
+    "Connection",
+    "NEED_DATA",
+    "PAUSED",
+    "ConnectionClosed",
+    "Data",
+    "EndOfMessage",
+    "Event",
+    "InformationalResponse",
+    "Request",
+    "Response",
+    "CLIENT",
+    "CLOSED",
+    "DONE",
+    "ERROR",
+    "IDLE",
+    "MUST_CLOSE",
+    "SEND_BODY",
+    "SEND_RESPONSE",
+    "SERVER",
+    "SWITCHED_PROTOCOL",
+    "ProtocolError",
+    "LocalProtocolError",
+    "RemoteProtocolError",
+)
--- a/h11/_connection.py
+++ b/h11/_connection.py
@@ -1,24 +1,49 @@
 # This contains the main Connection class. Everything in h11 revolves around
 # this.
+from typing import Any, Callable, cast, Dict, List, Optional, Tuple, Type, Union
 
-from ._events import *  # Import all event types
+from ._events import (
+    ConnectionClosed,
+    Data,
+    EndOfMessage,
+    Event,
+    InformationalResponse,
+    Request,
+    Response,
+)
 from ._headers import get_comma_header, has_expect_100_continue, set_comma_header
-from ._readers import READERS
+from ._readers import READERS, ReadersType
 from ._receivebuffer import ReceiveBuffer
-from ._state import *  # Import all state sentinels
-from ._state import _SWITCH_CONNECT, _SWITCH_UPGRADE, ConnectionState
+from ._state import (
+    _SWITCH_CONNECT,
+    _SWITCH_UPGRADE,
+    CLIENT,
+    ConnectionState,
+    DONE,
+    ERROR,
+    MIGHT_SWITCH_PROTOCOL,
+    SEND_BODY,
+    SERVER,
+    SWITCHED_PROTOCOL,
+)
 from ._util import (  # Import the internal things we need
     LocalProtocolError,
-    make_sentinel,
     RemoteProtocolError,
+    Sentinel,
 )
-from ._writers import WRITERS
+from ._writers import WRITERS, WritersType
 
 # Everything in __all__ gets re-exported as part of the h11 public API.
 __all__ = ["Connection", "NEED_DATA", "PAUSED"]
 
-NEED_DATA = make_sentinel("NEED_DATA")
-PAUSED = make_sentinel("PAUSED")
+
+class NEED_DATA(Sentinel, metaclass=Sentinel):
+    pass
+
+
+class PAUSED(Sentinel, metaclass=Sentinel):
+    pass
+
 
 # If we ever have this much buffered without it making a complete parseable
 # event, we error out. The only time we really buffer is when reading the
@@ -44,7 +69,7 @@ DEFAULT_MAX_INCOMPLETE_EVENT_SIZE = 16 * 1024
 # our rule is:
 # - If someone says Connection: close, we will close
 # - If someone uses HTTP/1.0, we will close.
-def _keep_alive(event):
+def _keep_alive(event: Union[Request, Response]) -> bool:
     connection = get_comma_header(event.headers, b"connection")
     if b"close" in connection:
         return False
@@ -53,7 +78,9 @@ def _keep_alive(event):
     return True
 
 
-def _body_framing(request_method, event):
+def _body_framing(
+    request_method: bytes, event: Union[Request, Response]
+) -> Tuple[str, Union[Tuple[()], Tuple[int]]]:
     # Called when we enter SEND_BODY to figure out framing information for
     # this body.
     #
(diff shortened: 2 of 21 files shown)


h11/__init__.py
```diff
@@ -6,16 +6,57 @@
...
+from h11._version import __version__
 
 PRODUCT_ID = "python-h11/" + __version__
 
 
-__all__ = ["ProtocolError", "LocalProtocolError", "RemoteProtocolError"]
-__all__ += _events.__all__
-__all__ += _connection.__all__
-__all__ += _state.__all__
+__all__ = (
```
sethmlarson, line 38: This change already is going to do a ton for discoverability :tada: IDEs don't like dynamic `__all__`.

h11/_connection.py
```diff
@@ -243,15 +267,16 @@ def _process_event(self, role, event):
 
         # Then perform the updates triggered by it.
 
-        # self._request_method
-        if type(event) is Request:
+        if isinstance(event, Request):
```
sethmlarson, line 270: Is this an open Mypy bug? If not would be worth filing, `type(x) is y` is faster than `isinstance()`

h11/_headers.py
```diff
@@ -90,30 +93,39 @@ class Headers(Sequence):
...
         _, name, value = self._full_items[idx]
         return (name, value)
 
-    def raw_items(self):
+    def raw_items(self) -> List[Tuple[bytes, bytes]]:
         return [(raw_name, value) for raw_name, _, value in self._full_items]
 
 
-def normalize_and_validate(headers, _parsed=False):
+HeaderTypes = Union[
```
sethmlarson, line 119: Shouldn't this be `List[Tuple[Union[str, bytes], Union[str, bytes]]` instead of the individual permutations?

h11/_headers.py
```diff
@@ -90,30 +93,39 @@ class Headers(Sequence):
...
+    List[Tuple[bytes, bytes]],
+    List[Tuple[bytes, str]],
+    List[Tuple[str, bytes]],
+    List[Tuple[str, str]],
+    Headers,
+]
+
+
+def normalize_and_validate(headers: HeaderTypes, _parsed: bool = False) -> Headers:
```
sethmlarson, line 128: Thoughts on using `@overload` here with `_parsed=True` only accepting `Headers | List[Tuple[bytes, bytes]]`? Might even remove the need for the `assert bytes`?

h11/_readers.py
```diff
@@ -32,9 +54,9 @@
 obs_fold_re = re.compile(br"[ \t]+")
 
 
-def _obsolete_line_fold(lines):
+def _obsolete_line_fold(lines: Iterable[bytes]) -> Generator[bytes, None, None]:
```
sethmlarson, line 57: Is `Generator` relevant to this API since we're not using `.send()`? Maybe should be `Iterable[bytes]`?

h11/_readers.py
```diff
@@ -52,7 +74,9 @@ def _obsolete_line_fold(lines):
         yield last
 
 
-def _decode_header_lines(lines):
+def _decode_header_lines(
+    lines: Iterable[bytes],
+) -> Generator[Tuple[bytes, bytes], None, None]:
```
sethmlarson, line 79: Same as above.

h11/_writers.py
```diff
@@ -7,14 +7,17 @@
...
+from ._headers import Headers
 from ._state import CLIENT, IDLE, SEND_BODY, SEND_RESPONSE, SERVER
-from ._util import LocalProtocolError
+from ._util import LocalProtocolError, Sentinel
 
 __all__ = ["WRITERS"]
 
 
-def write_headers(headers, write):
+def write_headers(headers: Headers, write: Callable[[bytes], None]) -> None:
```
sethmlarson, line 20: Since we don't care about the return type of `write` should be `Callable[[bytes], Any]`?

h11/_writers.py
```diff
@@ -28,15 +31,17 @@ def write_headers(headers, write):
     write(b"\r\n")
 
 
-def write_request(request, write):
+def write_request(request: Request, write: Callable[[bytes], None]) -> None:
```
sethmlarson, line 34: Same here and below

sethmlarson (commented):
Overall this looks good. What are your thoughts on having the edges of the API be fully type-hinted but using `# type: ignore[...]` more liberally within APIs to not require using `assert`, `isinstance`, etc? Wondering if these changes would impact performance.

h11/_connection.py
pgjones: https://mypy.readthedocs.io/en/stable/common_issues.html#complex-type-tests

Bit of a pain - we could drop an assert in though rather than use a cast?

h11/_headers.py
pgjones: mypy seems to get confused by that (although technically it is more correct as it doesn't need to uniformly the same type in the tuple for all elements of the list) - I think this covers enough of the likely use cases and mypy understands it.

h11/_headers.py
pgjones: :) - yea, I'll try that again - I couldn't get mypy to figure out what I meant when I tried, see above comment as well.

h11/_readers.py
pgjones: Yea, changed. Thanks

h11/_writers.py
pgjones: Yea, changed. Thanks

h11/_writers.py
pgjones: Yea, changed. Thanks

h11/_readers.py
pgjones: Yea, changed. Thanks

pgjones:
> What are your thoughts on having the edges of the API be fully type-hinted but using # type: ignore[...] more liberally within APIs 

I'm hoping to use mypyc next to gain more performance (it was reasonably effective 2 years ago). So I'm not keen on this unless it doesn't affect mypyc.

h11/_connection.py
pgjones: Fixed, with only a couple of casts.

h11/_headers.py
pgjones: Done.

h11/_util.py
```diff
@@ -97,21 +101,21 @@ def validate(regex, data, msg="malformed data", *format_args):
...
-    def __repr__(self):
+class Sentinel(type):
+    def __repr__(self) -> str:
         return self.__name__
 
 
-def make_sentinel(name):
-    cls = _SentinelBase(name, (_SentinelBase,), {})
-    cls.__class__ = cls
+def make_sentinel(name: str) -> Sentinel:
```
graingert, line 109: how about this?

```python
from typing import TypeVar, Type, Tuple, Any, Dict

_T_Sentinel = TypeVar("_T_Sentinel", bound="Sentinel")


class Sentinel(type):
    def __new__(
        cls: Type[_T_Sentinel],
        name: str,
        bases: Tuple[type, ...],
        namespace: Dict[str, Any],
        **kwds: Any
    ) -> _T_Sentinel:
        assert bases == (Sentinel, )
        v = super().__new__(cls, name, bases, namespace, **kwds)
        v.__class__ = v  # type: ignore
        return v

    def __repr__(self) -> str:
        return self.__name__


class CLIENT(Sentinel, metaclass=Sentinel):
    pass
```

pgjones: Looks great, thanks!

graingert: @pgjones you marked this as resolved - but I don't think it's reflected in the PR?

pgjones: Was missing in the commit, now added.

h11/_state.py
```diff
@@ -129,26 +130,70 @@
     "ERROR",
 ]
 
-CLIENT = make_sentinel("CLIENT")
-SERVER = make_sentinel("SERVER")
+
+class CLIENT(Sentinel, metaclass=Sentinel):
+    pass
```
tomchristie, line 135: I'm surprised that `metaclass=...` is needed here.

(I guess I would have assumed that inheriting from a class with a custom `__new__` method was sufficient.)

Not sure this is an actionable comment, but just "this is a bit fiddly, innit?" 🙂

h11/_readers.py
```diff
@@ -89,22 +104,29 @@ def maybe_read_from_SEND_RESPONSE_server(buf):
...
         raise LocalProtocolError("no response line received")
     matches = validate(status_line_re, lines[0], "illegal status line: {!r}", lines[0])
-    # Tolerate missing reason phrases
-    if matches["reason"] is None:
-        matches["reason"] = b""
-    status_code = matches["status_code"] = int(matches["status_code"])
-    class_ = InformationalResponse if status_code < 200 else Response
+    http_version = (
+        b"1.1" if matches["http_version"] is None else matches["http_version"]
+    )
```
tomchristie, line 109: This is an interesting change. Much neater/more obvious this way, now that we're not passing an open-ended `**matches` to the response class.

h11/_readers.py
```diff
@@ -113,7 +135,7 @@ def __call__(self, buf):
         self._remaining -= len(data)
         return Data(data=data)
 
-    def read_eof(self):
+    def read_eof(self) -> NoReturn:
```
tomchristie, line 138: TIL: `typing.NoReturn`

h11/tests/test_helpers.py
```diff
@@ -1,12 +1,21 @@
...
 
 
-def test_normalize_data_events():
+def test_normalize_data_events() -> None:
     assert normalize_data_events(
         [
             Data(data=bytearray(b"1")),
             Data(data=b"2"),
-            Response(status_code=200, headers=[]),
+            Response(status_code=200, headers=[]),  # type: ignore[arg-type]
```
tomchristie, line 18: I'm curious about these `ignore` cases, since it looks correct to me?

h11/_writers.py
```diff
@@ -95,23 +108,32 @@ def send_data(self, data, write):
...
 
-WRITERS = {
+WritersType = Dict[
+    Union[Tuple[Sentinel, Sentinel], Sentinel],
+    Union[
+        Dict[str, Type[BodyWriter]],
+        Callable[[Union[InformationalResponse, Response], Writer], None],
+        Callable[[Request, Writer], None],
+    ],
+]
```
tomchristie, line 134: I think this typing here helps illuminate that the `WRITERS` usage here is a bit odd, with it's special case for `SEND_BODY`. 😀
Not actionable tho - just a passing observation.

h11/_headers.py
```diff
@@ -90,30 +99,57 @@ class Headers(Sequence):
...
+
+
+@overload
+def normalize_and_validate(headers: Headers, _parsed: Literal[True]) -> Headers:
+    ...
+
+
+@overload
+def normalize_and_validate(headers: HeaderTypes, _parsed: Literal[False]) -> Headers:
+    ...
```
tomchristie, line 140: I don't understand this overloading?
What's with the `Literal[True]`, `Literal[False]`, `bool` cases?

tomchristie (approved):
Reviewed as best as possible.
Pretty substantial bit of work here. 👍

A few passing comments along the way but no obvious blockers.

h11/_state.py
pgjones: Yea it is. There is some extra discussion [here](https://github.com/python-hyper/h11/pull/135#discussion_r712092061).

h11/tests/test_helpers.py
pgjones: I think mypy gets confused about the type of the empty list. There is an open bug I can't find at the moment

h11/_headers.py
pgjones: It allows mypy to determine that the type of `headers` changes depending on `_parsed` being True or False. See also [this](https://github.com/python-hyper/h11/pull/135#discussion_r697890517)

pgjones:
Thanks, I'll merge tomorrow

pgjones:
Thanks again for the reviews and comments 😄

PR #136, opened 2021-08-31
Fix typo on basic usage

Kludex:


Kludex:
Sorry, I was tired when I read that. 😅

PR #138, opened 2021-12-19
Change supported Python versions

pgjones:
Python 3.6 is at EOL and 3.10 was released around 2 months ago. This
allows the conditional dataclasses requirement to be removed.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,11 +13,12 @@ jobs:
       max-parallel: 5
       matrix:
         python-version:
-        - 3.6
         - 3.7
         - 3.8
         - 3.9
-        - pypy3
+        - "3.10"
+        - pypy-3.7
+        - pypy-3.8
 
     steps:
     - uses: actions/checkout@v2
--- a/README.rst
+++ b/README.rst
@@ -112,7 +112,7 @@ library.
 It has a test suite with 100.0% coverage for both statements and
 branches.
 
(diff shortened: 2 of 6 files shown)


pgjones:
Looks like the pypy3 gh action version is 3.6, rather than 3.7 or higher.

pgjones:
> Looks like the pypy3 gh action version is 3.6, rather than 3.7 or higher.

Now fixed by being explicit.

sethmlarson (approved):
Sounds good to me!

tomchristie:
Probably reasonable, yeah.

I guess a couple of questions here...

* What are `h11`'s biggest dependants, and when are they planning on dropping 3.6? (I'd assume `hypercorn`, `uvicorn`, `httpx`? Am I missing any big ones there? We haven't planned on dropping 3.6 for `httpx` and `uvicorn` yet, but also that *doesn't* necessarily mean that `h11` shouldn't/couldn't.)
* What are the usage figures for 3.6 atm? (Perhaps it doesn't much matter. I suppose [pypistats](https://pypistats.org/packages/h11) might be the best resource here? Indicates ~5%)
* I was going to ask what the best resource was for Python version lifespans, but I think [this answers that question](https://www.python.org/downloads/). I'm curious what state 3.6 is going to be marked as once `2021-12-23` ticks over.

pgjones:
Python3.6 is now long pass it's end of life and 3.11 is due soon, time to merge.

PR #142, opened 2022-01-20
GitHub action for automated deployment

tomchristie:
Hi folks 👋🏼

Here's a pass at what I think we'd need in order to have automated deployments, triggered by tagging a GitHub release.

* A PyPI API token would [need to be created for the package](https://pypi.org/help/#apitoken).
* The API token then needs to be [added as `PYPI_TOKEN` in the repo settings](https://github.com/python-hyper/h11/settings/secrets/actions).

The PR would perhaps also need to be fleshed out with something like a `DEPLOYMENT.md` text (or similar?) to describe the release process.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -1,0 +1,30 @@
+name: Publish
+
+on:
+  push:
+    tags:
+      - 'v*'
+
+jobs:
+  publish:
+    name: "Publish release"
+    runs-on: "ubuntu-latest"
+
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-python@v2
+        with:
+          python-version: 3.10
+      - name: "Install dependancies"
+        run: |
+          python -m pip install --upgrade pip setuptools
+          pip install --upgrade twine wheel
+      - name: "Build package"
+        run: |
+          python setup.py sdist bdist_wheel
+      - name: "Publish to PyPI"
+        run: |
+          twine upload dist/*
+        env:
+          TWINE_USERNAME: __token__
+          TWINE_PASSWORD: ${{ secrets.PYPI_TOKEN }}


sethmlarson:
Is python-hyper's policy of auto-inviting anyone who commits to have the commit bit compatible with tag-to-deploy release strategy? All it'd take is one account to go rogue and we'd have a compromised package.

tomchristie:
Yes - who gets the publish keys is a pretty tricky problem, isn't it.

This would be easier if GitHub's policy happened to be that ["Maintain"-level access](https://docs.github.com/en/organizations/managing-access-to-your-organizations-repositories/repository-roles-for-an-organization) was required in order to create releases. That'd be helpful because you could have the majority of a team with "Write"-level access, but then just 2 or 3 folks with "Maintain" or "Admin", who're able to hit the big red deploy button.

I've taken a big dig around GitHub's docs on securing actions etc. to see if there's a good solution to this at the moment. Perhaps there's some creative way around, but I can't find anything that's quite suitable. Which is a shame, because GitHub Actions + releases together are *otherwise* a really fantastic way of managing deploys. (And are great if you *are* okay with a policy of write-access maintainers can deploy.)

It looks like GitHub enterprise [provides custom roles](https://docs.github.com/en/enterprise-cloud@latest/organizations/managing-peoples-access-to-your-organization-with-roles/managing-custom-repository-roles-for-an-organization#custom-role-examples), currently in beta, which *I think would* allow an appropriate setup. (General team with PR read/write permissions, plus a small admin team with release management permissions.)

Those(non-existent) [custom actions are visible in `h11`'s settings](https://github.com/python-hyper/h11/settings/role_details) but there's no "create a role" button.

tomchristie:
Right - there *is* a way you can do this, while still having admin sign-off on deploys.

Create a [GitHub environment](https://www.aaron-powell.com/posts/2021-01-11-using-environments-for-approval-workflows-with-github/) with required reviews. You can then add the `PYPI_TOKEN` to that environment only.

You can then specify in the workflow config to run against that environment.

Issue #133, opened 2021-07-06
Robust handling of whitespace-malformed chunk encoding headers.


tomchristie:
Prompted by https://github.com/encode/httpx/discussions/1735

When a server uses chunked encoding, but erroneously includes whitespace between the `<chunk size marker>` and the `<crlf>`, a `RemoteProtocolError` is raised...

```pycon
>>> import h11
>>> state = h11.Connection(our_role=h11.CLIENT)
>>> state.send(h11.Request(method=b'GET', target='b/', headers=[(b'Host', 'example.com')]))
b'GET b/ HTTP/1.1\r\nHost: example.com\r\n\r\n'
>>> state.send(h11.EndOfMessage())
b''
>>> state.receive_data(b'HTTP/1.1 200 OK\r\n')
>>> state.receive_data(b'Transfer-Encoding: chunked\r\n')
>>> state.receive_data(b'\r\n')
>>> state.receive_data(b'123  \r\n')
>>> state.receive_data(b'x' * 123)
>>> state.receive_data(b'\r\n0\r\n\r\n')
>>> state.next_event()
Response(status_code=200, headers=<Headers([(b'transfer-encoding', b'chunked')])>, http_version=b'1.1', reason=b'OK')
>>> state.next_event()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "venv/lib/python3.8/site-packages/h11/_connection.py", line 443, in next_event
    exc._reraise_as_remote_protocol_error()
  File "venv/lib/python3.8/site-packages/h11/_util.py", line 76, in _reraise_as_remote_protocol_error
    raise self
  File "venv/lib/python3.8/site-packages/h11/_connection.py", line 425, in next_event
    event = self._extract_next_receive_event()
  File "/venv/lib/python3.8/site-packages/h11/_connection.py", line 367, in _extract_next_receive_event
    event = self._reader(self._receive_buffer)
  File "venv/lib/python3.8/site-packages/h11/_readers.py", line 157, in __call__
    matches = validate(
  File "venv/lib/python3.8/site-packages/h11/_util.py", line 88, in validate
    raise LocalProtocolError(msg)
h11._util.RemoteProtocolError: illegal chunk header: bytearray(b'123  \r\n')
```

No doubt this is technically correct, but we *probably* want to be robust to this kind of malformed input, given that we've seen it occur as a user-issue in `httpx`.

njsmith:
[...]

PR #143, opened 2022-01-26
Better handle of trailing spaces on chunk header - Fixes issue #133

bratao:
Hello,

I found some servers that return multiple trailing spaces on the chunk header. I changed the regex to match on those cases. This fixes issue #133

### DIFF
--- a/h11/_abnf.py
+++ b/h11/_abnf.py
@@ -125,5 +125,5 @@ chunk_ext = r";.*"
 chunk_header = (
     r"(?P<chunk_size>{chunk_size})"
     r"(?P<chunk_ext>{chunk_ext})?"
-    r"\r\n".format(**globals())
+    r"\s*\r\n".format(**globals())
 )
--- a/h11/_abnf.py
+++ b/h11/_abnf.py
@@ -125,5 +125,6 @@ chunk_ext = r";.*"
 chunk_header = (
     r"(?P<chunk_size>{chunk_size})"
     r"(?P<chunk_ext>{chunk_ext})?"
-    r"{OWS}\r\n".format(**globals())
+    r"{OWS}\r\n".format(**globals())  # Even though the specification does not allow for extra whitespaces, 
+    # we are lenient with trailing whitespaces because some servers on the wild use it.
 )
--- a/h11/_abnf.py
+++ b/h11/_abnf.py
@@ -125,5 +125,5 @@ chunk_ext = r";.*"
 chunk_header = (
     r"(?P<chunk_size>{chunk_size})"
     r"(?P<chunk_ext>{chunk_ext})?"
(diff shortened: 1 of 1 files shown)


h11/_abnf.py
```diff
@@ -125,5 +125,5 @@
 chunk_header = (
     r"(?P<chunk_size>{chunk_size})"
     r"(?P<chunk_ext>{chunk_ext})?"
-    r"\r\n".format(**globals())
+    r"\s*\r\n".format(**globals())
```
pgjones, line 128: I think this should be `OWS` rather than `\s`, especially as I think the latter allows CRLF

bratao: Done!

njsmith:
Please also add a comment noting how this production deviates from the standard and why, and also a test.

bratao:
Following https://datatracker.ietf.org/doc/html/rfc7230#section-3.3.1 , thje grammar is the following:

```
  The chunked transfer coding wraps the payload body in order to
   transfer it as a series of chunks, each with its own size indicator,
   followed by an OPTIONAL trailer containing header fields.  Chunked
   enables content streams of unknown size to be transferred as a
   sequence of length-delimited buffers, which enables the sender to
   retain connection persistence and the recipient to know when it has
   received the entire message.

     chunked-body   = *chunk
                      last-chunk
                      trailer-part
                      CRLF

     chunk          = chunk-size [ chunk-ext ] CRLF
                      chunk-data CRLF
     chunk-size     = 1*HEXDIG
     last-chunk     = 1*("0") [ chunk-ext ] CRLF

     chunk-data     = 1*OCTET ; a sequence of chunk-size octets

   The chunk-size field is a string of hex digits indicating the size of
   the chunk-data in octets.  The chunked transfer coding is complete
   when a chunk with a chunk-size of zero is received, possibly followed
   by a trailer, and finally terminated by an empty line.

   A recipient MUST be able to parse and decode the chunked transfer
   coding.
```


I cannot see why a server would add an empty space on this part. Maybe as a bad coded placeholder for [ chunk-ext ].

But I seen this on the wild ( and #133), and I cannot image a security problem or any kind of request smuggling by skipping the white spaces.

njsmith:
Sorry, I meant a `# comment` in the source code :-) So the next person reading the code isn't confused about why the production in the code doesn't match the one in the spec.

bratao:
@njsmith Oh, now I get it. See if this comment fits it.

mariuspodean:
Hi guys, thank for the fix! Any ideas when this will get merged?

bratao:
@njsmith do you need me to do anything else to get this upstream?

pgjones:
I've merged this manually and added a test.

PR #145, opened 2022-03-26
Reformat with black 22.1.0

madkinsz:
This is the latest version of black and is required for new commits to pass CI.


h11/_readers.py
```diff
@@ -40,7 +40,7 @@
 
 # Remember that this has to run in O(n) time -- so e.g. the bytearray cast is
 # critical.
-obs_fold_re = re.compile(br"[ \t]+")
+obs_fold_re = re.compile(rb"[ \t]+")
```
madkinsz, line 43: Out of curiosity, why is it not O(n) without the byte array cast?

pquentin: This comment is confusing, it should be placed just before those lines:

```
            if not isinstance(last, bytearray):
                last = bytearray(last)
```

[bytearray](https://docs.python.org/3/library/stdtypes.html#bytearray) is mutable. With normal immutable `bytes`, each append creates a new copy, so this would become accidentally quadratic.

madkinsz: That makes way more sense. https://github.com/python-hyper/h11/pull/146

sethmlarson (approved):

PR #148, opened 2022-04-25
Use `email.utils.format_datetime` instead of undocumented `wsgiref.handlers.format_date_time`

GalaxySnail:
Remove an undocumented function call in `examples/trio-server.py`.


pgjones:
Thanks

PR #149, opened 2022-04-25
Call `conn.send_failed` if `stream.send_all` fails with an exception.

GalaxySnail:
See details:

[1] https://trio.readthedocs.io/en/latest/reference-io.html#trio.abc.SendStream.send_all
[2] https://h11.readthedocs.io/en/latest/api.html#h11.Connection.send_failed


pgjones:
Thanks

PR #150, opened 2022-04-25
Remove python version in trio-server.py

GalaxySnail:
Now trio 0.20.0 requires python 3.7 or later. But IMO it's not necessary to mention it here and update it every year.

### DIFF
--- a/examples/trio-server.py
+++ b/examples/trio-server.py
@@ -1,6 +1,5 @@
 # A simple HTTP server implemented using h11 and Trio:
 #   http://trio.readthedocs.io/en/latest/index.html
-# (so requires python 3.5+).
 #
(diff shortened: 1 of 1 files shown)


pquentin (approved):
To be clear is currently Trio is 3.6+, just like h11.

GalaxySnail:
I'm not sure if I missed something, but [trio 0.20.0](https://pypi.org/project/trio/0.20.0/#data) which was released on 2022-02-21 requires Python 3.7+ and python 3.6 support was dropped in [GH-2210](https://github.com/python-trio/trio/pull/2210).

pquentin:
You're right, sorry.

pgjones:
Thanks

PR #154, opened 2022-08-25
Switch from Sentinel types to Enums

### LINKED ISSUE #153: provide a new version of Connection that returns/accepts enum types for PAUSED/NEED_DATA
the current Connection.next_event `Event | type[_Sentinel]` return type is [slightly painful to work with](https://github.com/encode/httpcore/pull/526/files#r901148647). It would be useful to have a version that returns enum based sentinels instead

pgjones:
The latter are much easier to work with when type hinting and can be
used successfully with mypyc, whereas the former are sadly very
difficult in both aspects.

This loses the nice property of `type(NEED_DATA) is NEED_DATA` (as
expanded on in the deleted docs section). However, I don't think this
is widely used in practice.

Closes #153. 

@njsmith what do you think about this?


tomchristie:
This looks to me like a breaking API change that would impact [Uvicorn](https://github.com/encode/uvicorn/blob/f460f7ade7eff84b5017b7f52c9f1cdcb5c8b769/uvicorn/protocols/http/h11_impl.py#L173) and [HTTPX](https://github.com/encode/httpcore/blob/225a067ee5727b7468c1ad4bf9e954cc0ab7cfc9/httpcore/_async/http11.py#L168).

I do prefer the API, tho.

Kludex:
I can check this on Uvicorn tonight. jfyk

tomchristie:
@Kludex Sure thing. Permalinks to relevant parts of code in the comment above. (Tho I can see that's not obvious.)

Kludex:
This was merged on `httpcore`: https://github.com/encode/httpcore/pull/579

Kludex:
The only problem on `uvicorn` was:
```python
            event = h11.InformationalResponse(
                status_code=100, headers=[], reason="Continue"
            )
```

Issue:
```bash
uvicorn/protocols/http/h11_impl.py:537: error: Argument "headers" to "InformationalResponse" has incompatible type "List[<nothing>]"; expected "Union[Headers, List[Tuple[bytes, bytes]], List[Tuple[str, str]]]"  [arg-type]
uvicorn/protocols/http/h11_impl.py:537: note: "List" is invariant -- see https://mypy.readthedocs.io/en/stable/common_issues.html#variance
uvicorn/protocols/http/h11_impl.py:537: note: Consider using "Sequence" instead, which is covariant
Found 1 error in 1 file (checked 53 source files)
```

I've used this branch on `uvicorn` to check the changes I'd need: https://github.com/encode/uvicorn/pull/1744/files

PR #157, opened 2022-12-24
Don't set connection close headers when they are already set on the Response

Kludex:
- Closes https://github.com/python-hyper/h11/issues/156


Kludex:
This cache is extremely painful.

Issue #158, opened 2023-04-02
Remove test files when publishing package


dosisod:
I was poking around my `.venv` folder and noticed h11 includes the `test/` folder in it's published packages. Is this intentional? The `setup.py` and `MANIFEST.in` file seem to indicate that the tests should be added, but I don't see any explanation as to why:

https://github.com/python-hyper/h11/blob/cdccbeff44a39426b58010eb454a75015ec6f8bc/setup.py#L18-L20

and in the manifest file:

https://github.com/python-hyper/h11/blob/cdccbeff44a39426b58010eb454a75015ec6f8bc/MANIFEST.in#L3

Normally tests are only for development purposes and have no use to the client, so including the test folder just takes up space. Unless I am missing something, I think we can remove this folder (I wouldn't mind opening a PR for this, just though I would ask before doing so).

GalaxySnail:
It may not be a good idea to remove tests (and also documentations) from sdists. Here are some reasons:

- https://discuss.python.org/t/should-sdists-include-docs-and-tests/14578/115
- https://discuss.python.org/t/sdists-for-pure-python-projects/25191/10

IMO it should be fine to remove tests from wheels and keep them in sdists, but I don't know how to do it.

dosisod:
I agree with that tests and docs should stay in the sdists. I haven't double checked this, but I believe the `setup.py` file dictates what files are included in the wheel files, and the `MANIFEST.in` file dictates what goes in the sdist (according to this [post on stack overflow](https://stackoverflow.com/questions/26911045/is-it-possible-to-exclude-certain-files-when-building-a-wheel-with-setup-py)).

I assume that the `include_package_data` flag in the `setup.py` file is needed to include the `py.typed` file, so perhaps we can configure [`exclude_package_data`](https://setuptools.pypa.io/en/latest/userguide/datafiles.html#exclude-package-data) to exclude the `h11/test` folder?

PR #159, opened 2023-04-14
Remove `tests` folder from wheels

dosisod:
I have verified on my local machine that the source distributions still include the tests folder. Doing so reduces the zipped wheel file from 57KB to 37KB, which is about 35% smaller.

Closes https://github.com/python-hyper/h11/issues/158.


setup.py
```diff
@@ -12,12 +12,9 @@
     author="Nathaniel J. Smith",
     author_email="njs@pobox.com",
     license="MIT",
-    packages=find_packages(),
+    packages=find_packages(exclude=["h11.tests"]),
     package_data={'h11': ['py.typed']},
     url="https://github.com/python-hyper/h11",
-    # This means, just install *everything* you see under h11/, even if it
-    # doesn't look like a source file, so long as it appears in MANIFEST.in:
-    include_package_data=True,
```
sigmavirus24, line 20: Did you verify the sdist tarball has the tests? If not, please do as others rely on their presence there

sigmavirus24: 🤦missed that. Saw the comment about reducing artifact size and was pretty sure this boolean disabled MANIFEST.in but maybe tooling has improved to not conflate the two.

pgjones (approved):

zanieb (approved):

sethmlarson (approved):

sethmlarson (approved):

dosisod:
Thanks!

PR #162, opened 2023-06-08
Format with black 23.3.0

zanieb:
The `black` version used in CI is not pinned so the new version's formatting needs to be applied to `master`.

PR #163, opened 2023-06-08
Pin format requirements for the project

zanieb:
Includes https://github.com/python-hyper/h11/pull/162

Pins formatting dependencies and moves definition into a `format-requirements.txt` so it can be updated automatically.

We may want to do something similar for `mypy`.


sethmlarson (approved):
LGTM

PR #171, opened 2024-03-02
Don't set connection close headers when they are already set on the Response

Kludex:
- Closes https://github.com/python-hyper/h11/issues/156

This needs:
- [ ] https://github.com/python-hyper/h11/pull/172


Kludex:
It's not the cache... How is master passing?

Kludex:
@sethmlarson 👀 🙏 

Also, https://github.com/python-hyper/h11/pull/173 should be merged, I guess...

Kludex:
I think I deleted my fork, and this was closed...

PR #180, opened 2025-01-09
Add headers.get and headers.getlist

Julien00859:
Hello there, first time contributing :)

I wanted to access a header value with little code, like `req.headers[b"content-type"]` but since the headers are implemented as a `Sequence` there are no easy way beside iterating all the headers to find the one that interest me.

I am not challenging the internal data-structure, `list(tuple(bytes, bytes, bytes))` is kinda the only sane structure because of headers like `Set-Cookie`.

In this work I add two helpers `get(name) -> bytes` and `getlist(name) -> list[bytes]` that searches the headers for me, so I don't have to write this `next(generator, None)` everywhere :smile: The two method names at coined at the `Headers` class of [`werkzeug`](https://werkzeug.palletsprojects.com/en/stable/datastructures/#werkzeug.datastructures.Headers.getlist). We might also add the `type=` parameter for total symmetry, you tell me.

The implementation is simple stupid, it iterates over the headers looking for one that matches the name.

I did experiment with another implementation, as follow (didn't test it):

```diff
     def __init__(self, full_items: List[Tuple[bytes, bytes, bytes]]) -> None:
         self._full_items = full_items
+        self._index = {}
+        for idx, (_, name, value) in enumerate(full_items):
+            self._index.setdefault(name, []).append(idx)
 
     def __bool__(self) -> bool:
         return bool(self._full_items)
@@ -118,6 +133,15 @@ class Headers(Sequence[Tuple[bytes, bytes]]):
         _, name, value = self._full_items[idx]
         return (name, value)
 
+    def get(self, name: bytes, default: T = None) -> Union[bytes, T]:
+        try:
+            self[self._index[name][0]][1]
+        except KeyError:
+            return default
+
+    def getlist(self, name: bytes) -> List[bytes]:
+        return [self[idx][1] for idx in self._index.get(name, ())]
```

But I don't think it deserves an index. I think the O(n) search is fine.

### DIFF
--- a/h11/_headers.py
+++ b/h11/_headers.py
@@ -1,5 +1,5 @@
 import re
-from typing import AnyStr, cast, List, overload, Sequence, Tuple, TYPE_CHECKING, Union
+from typing import List, overload, Sequence, Tuple, TYPE_CHECKING, TypeVar, Union
 
 from ._abnf import field_name, field_value
 from ._util import bytesify, LocalProtocolError, validate
@@ -12,6 +12,8 @@ try:
 except ImportError:
     from typing_extensions import Literal  # type: ignore
 
+
+T = TypeVar("T")
 
 # Facts
 # -----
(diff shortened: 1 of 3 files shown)


h11/_headers.py
```diff
@@ -1,5 +1,5 @@
 import re
-from typing import AnyStr, cast, List, overload, Sequence, Tuple, TYPE_CHECKING, Union
+from typing import List, overload, Sequence, Tuple, TYPE_CHECKING, TypeVar, Union
```
Julien00859, line 2: My IDE was telling me `AnyStr` and `cast` were not used, I used the opportunity of changing the line to remove them

Julien00859:
Hello @njsmith, do you consider this work valuable? If not would you like to close this PR?

PR #184, opened 2025-07-05
Fix `test_max_incomplete_event_size_countermeasure` test

anton-ryzhov:
This test was not actually testing what it should — `max_incomplete_event_size` limit is only checked for incomplete events, for request before `"\r\n\r\n"`.

`get_all_events()` before `"\r\n\r\n"` conditionally raises an exception depending on `max_incomplete_event_size`

### DIFF
--- a/h11/tests/test_connection.py
+++ b/h11/tests/test_connection.py
@@ -425,6 +425,8 @@ def test_max_incomplete_event_size_countermeasure() -> None:
     c = Connection(SERVER, max_incomplete_event_size=5000)
     c.receive_data(b"GET / HTTP/1.0\r\nBig: ")
     c.receive_data(b"a" * 4000)
+    assert get_all_events(c) == []
+
     c.receive_data(b"\r\n\r\n")
     assert get_all_events(c) == [
         Request(
@@ -437,7 +439,7 @@ def test_max_incomplete_event_size_countermeasure() -> None:
     c.receive_data(b"GET / HTTP/1.0\r\nBig: ")
     c.receive_data(b"a" * 4000)
     with pytest.raises(RemoteProtocolError):
-        c.next_event()
+        get_all_events(c)
 
     # Temporarily exceeding the size limit is fine, as long as its done with
     # complete events:


anton-ryzhov:
@njsmith does this PR miss something or could it be merged?

njsmith:
I guess I'm not seeing anything wrong with this, but I also don't understand what the purpose of this is? What was it not testing before that it does test now?

anton-ryzhov:
[These](https://github.com/python-hyper/h11/blob/62c5068c971579d61fa1b55373390e12f25fd856/h11/tests/test_connection.py#L423-L434) [two](https://github.com/python-hyper/h11/blob/62c5068c971579d61fa1b55373390e12f25fd856/h11/tests/test_connection.py#L436-L440) blocks demonstrate and test that the same payload is handled fine with `max_incomplete_event_size=5000` but raises an error with `max_incomplete_event_size=4000`.

But in fact the first block doesn't ensure that, it doesn't fail if we set [``max_incomplete_event_size=1000``](https://github.com/python-hyper/h11/compare/62c5068c971579d61fa1b55373390e12f25fd856...anton-ryzhov:h11:7cbd20af14cedfdce78685c4c4e9b5ce97a29af7). But it [does fail](https://github.com/python-hyper/h11/compare/62c5068c971579d61fa1b55373390e12f25fd856...anton-ryzhov:h11:7a9a93698e31e0f656af23086a658272bd69e3ea) if we add an assert.

So, the successful case was always successful regardless of the configuration

njsmith:
Ohh I see yeah that makes sense for the first change, good catch. Still don't think I understand the motivation for the second change?

anton-ryzhov:
> Still don't think I understand the motivation for the second change?

To be honest I don't remember.
 * Maybe it was for sake of consistency, to check the same in both (line 428) cases
 * Maybe a semantic change, instead of "the very first event raises an error" to "at any step of processing, this error will be triggered"
 * Maybe it played some role in [this change](https://github.com/python-hyper/h11/pull/185)

Do you want me to revert it?

PR #185, opened 2025-07-05
Fix parsing limits

### LINKED ISSUE #182: Inconsistent parsing behavior
I want to report an inconsistent behavior we've observed. It affects production, was very confusing and hard to understand/debug.

Initially we had a replica set of `hypercorn`-based applications, and the same request was processed fine by the majority of instances, but not by one of a few others. The server didn't even log such requests.

After the investigation I've discovered that [this limit](https://github.com/python-hyper/h11/blob/70a96bea8e55403e5d92db14c111432c6d7a8685/h11/_connection.py#L69) is not always enforced. The library adds [next chunk](https://github.com/python-hyper/h11/blob/70a96bea8e55403e5d92db14c111432c6d7a8685/h11/_connection.py#L406) and tries to [parse it](https://github.com/python-hyper/h11/blob/70a96bea8e55403e5d92db14c111432c6d7a8685/h11/_connection.py#L481-L483) before [enforcing the limit](https://github.com/python-hyper/h11/blob/70a96bea8e55403e5d92db14c111432c6d7a8685/h11/_connection.py#L485). So if a request exceeds the limit, it will still be parsed successfully if it comes in one chunk (which is often the case).

So practically the same request could be parsed successfully or not depending on how the OS and TCP stack chunks it.

I think it should be more stable and consistent. Should the library check buffer size before the parsing attempt? What do you think?

Here is a demo:
[h11-issue-demo.zip](https://github.com/user-attachments/files/19753970/h11-issue-demo.zip)

```bash
$ pip install --requirement requirements.txt
$ pytest
FAILED test_client.py::test_fails_slightly_above_limit[uvicorn_endpoint-False] - assert not True
FAILED test_client.py::test_fails_slightly_above_limit[hypercorn_endpoint-False] - assert not True
========================================================================================== 2 failed, 10 passed in 2.16s ===========================================================================================
```

Inconsistency highlight:
```bash
$ pytest -k test_fails_slightly_above_limit
FAILED test_client.py::test_fails_slightly_above_limit[uvicorn_endpoint-False] - assert not True
FAILED test_client.py::test_fails_slightly_above_limit[hypercorn_endpoint-False] - assert not True
==================================================================================== 2 failed, 2 passed, 8 deselected in 0.84s ====================================================================================
```

anton-ryzhov:
Fixes #182 

`maybe_read_from_IDLE_client` and `maybe_read_from_SEND_RESPONSE_server` readers wait for concrete terminators and produce NEED_DATA events until that. After NEED_DATA `max_incomplete_event_size` limit is checked.

Here I'm adding consistent size limits for these events, that do not depend on chunking patterns.

To ensure that we need to reject  complete `request` data chunk if incomplete `request[:-1]` already exceeds `max_incomplete_event_size` limit.

Practically `max_complete_event_size = max_incomplete_event_size + 1` and enforced for "all or nothing" event readers.

I've added failing test with the first commit and the fix in the second.

### DIFF
--- a/h11/_connection.py
+++ b/h11/_connection.py
@@ -478,6 +478,7 @@ class Connection:
         if self.their_state is ERROR:
             raise RemoteProtocolError("Can't receive data when peer state is ERROR")
         try:
+            receive_buffer_len_before = len(self._receive_buffer)
             event = self._extract_next_receive_event()
             if event not in [NEED_DATA, PAUSED]:
                 self._process_event(self.their_role, cast(Event, event))
@@ -492,6 +493,15 @@ class Connection:
                     # We're still trying to complete some event, but that's
                     # never going to happen because no more data is coming
                     raise RemoteProtocolError("peer unexpectedly closed connection")
+
+            if isinstance(event, (Request, InformationalResponse, Response)):
+                read_from_buffer = receive_buffer_len_before - len(self._receive_buffer)
+                if read_from_buffer > self._max_incomplete_event_size + 1:
+                    # 431 is "Request header fields too large" which is pretty
+                    # much the only situation where we can get here
+                    raise RemoteProtocolError(
+                        "Receive buffer too long", error_status_hint=431
+                    )
             return event
         except BaseException as exc:
             self._process_error(self.their_role)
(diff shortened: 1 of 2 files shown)


anton-ryzhov:
@njsmith could you please take a look?

PR #187, opened 2025-09-14
Add support for Python 3.14 and drop EOL 3.8-3.9

hugovk:
Also test PyPy3.11, because the latest releases no longer contain PyPy3.9 or PyPy3.10:

https://pypy.org/posts/2025/07/pypy-v7320-release.html

(We could also drop Python 3.9, it's EOL next month: https://devguide.python.org/versions/)

Edit: 3.9 now also dropped.


h11/_connection.py
```diff
@@ -1,17 +1,6 @@
...
-    cast,
-    Dict,
-    List,
-    Optional,
-    overload,
-    Tuple,
-    Type,
-    Union,
-)
+from typing import Any, Callable, cast, Optional, overload, Union
```
sigmavirus24: I believe the desire is to reduce merge conflicts by keeping these in separate lines.

cclauss (approved):

PR #192, opened 2026-01-27
modernize the package setup

Vizonex:
I decided to start making contributions to the project in the hopes of starting to find new ways to optimize it since there are many larger libraries such as `uvicorn` and a few others that make use of this library. I wanted to start with something small which was to try and first modernize the package setup being used. It's not much, but anything that I could not solve I have decided on leaving that alone. 

I do however have a suggestion that I did not add yet because I have not seen the workflows for this project yet but would it be possible to move the tests directory from `h11/tests` a `tests` directory instead so that we don't have to worry about excluding it?


pyproject.toml
```diff
@@ -1,3 +1,42 @@
...
+
+[tool.setuptools.dynamic]
+version = {attr = "h11.__version__"}
+
+[project.urls]
+    homepage = "https://github.com/python-hyper/h11"
+    repository = "https://github.com/python-hyper/h11.git"
+
+
+
```
Kludex:
```suggestion
```

pyproject.toml
```diff
@@ -1,3 +1,42 @@
...
+[build-system]
+requires = ["setuptools"]
+
+[tool.setuptools.dynamic]
+version = {attr = "h11.__version__"}
+
+[project.urls]
+    homepage = "https://github.com/python-hyper/h11"
+    repository = "https://github.com/python-hyper/h11.git"
```
Kludex:
```suggestion
homepage = "https://github.com/python-hyper/h11"
repository = "https://github.com/python-hyper/h11.git"
```

Vizonex:
@Kludex thanks for the help.

PR #197, opened 2026-05-18
Broaden receive data buffer types

### LINKED ISSUE #186: Update types to adhere to `--strict-bytes` / PEP688
This is a near copy-paste of https://github.com/python-hyper/h2/issues/1305

Mypy 1.16 added [`--strict-bytes`](https://mypy.readthedocs.io/en/stable/command_line.html#cmdoption-mypy-strict-bytes) to `--strict`, and plans to enable it by default in mypy 2.0. As far as I can tell the type hints in this package has not been updated to reflect this, giving false alarms for downstream users.

Pyright also does not promote `bytearray`/`memoryview` to `bytes` with [`disableBytesTypePromotions`](https://microsoft.github.io/pyright/#/configuration?id=type-evaluation-settings) defaulting to `True`.

The one I noticed when updating https://github.com/pgjones/hypercorn/pull/303 was [`Connection.receive_data`](https://github.com/python-hyper/h11/blob/62c5068c971579d61fa1b55373390e12f25fd856/h11/_connection.py#L364) only accepting `bytes` for the `data` parameter, but it only passes `data` onto [`ReceiveBuffer.__iadd__`](https://github.com/python-hyper/h11/blob/62c5068c971579d61fa1b55373390e12f25fd856/h11/_receivebuffer.py#L53) which explicitly accepts `bytes | bytearray`. Even this could be widened, as [`bytearray.__iadd__`](https://github.com/python/typeshed/blob/c3a880b4282d86b5c27c81180827789c64af1eb4/stdlib/builtins.pyi#L841) effectively accepts `Buffer` (though I'm not sure if the [`ReadableBuffer` distinction](https://github.com/python/typeshed/blob/c3a880b4282d86b5c27c81180827789c64af1eb4/stdlib/_typeshed/__init__.pyi#L278) matters, probably not)

It's likely you can replace a lot of type hints with `collections.abc.Buffer`, or other protocols.

Mirochill:
## Summary
- widen `Connection.receive_data()` and the internal receive buffer annotations to accept `bytearray` and `memoryview` alongside `bytes`
- add a regression test covering both supported bytes-like inputs
- document the backward-compatible API expansion in the news fragment

## Why
Issue #186 points out that `receive_data()` already operates on bytes-like input, but its public annotation is still narrower than the runtime behavior. Widening the concrete accepted types removes false positives for downstream users adopting stricter bytes checking while keeping this patch intentionally small for h11's current typing/tooling baseline.

## Validation
- Not run locally
- Performed static review of the receive path and confirmed `Connection.receive_data()` forwards directly into `ReceiveBuffer.__iadd__()`.
- Verified the patch shape with `git diff --check`.
- Relying on the repository's remote CI for execution feedback.

Fixes #186.


h11/_connection.py
```diff
@@ -361,7 +361,7 @@ def trailing_data(self) -> Tuple[bytes, bool]:
         """
         return (bytes(self._receive_buffer), self._receive_buffer_closed)
 
-    def receive_data(self, data: bytes) -> None:
+    def receive_data(self, data: Union[bytes, bytearray, memoryview]) -> None:
```
GalaxySnail, line 364: It would be better to use `collections.abc.Buffer` for Python >= 3.12, as introduced in [PEP 688](https://peps.python.org/pep-0688/). Alternatively, you can use `if typing.TYPE_CHECKING: from typing_extensions import Buffer`, which avoids a runtime dependency on typing_extensions.

h11/_receivebuffer.py
```diff
@@ -50,7 +50,9 @@ def __init__(self) -> None:
         self._next_line_search = 0
         self._multiple_lines_search = 0
 
-    def __iadd__(self, byteslike: Union[bytes, bytearray]) -> "ReceiveBuffer":
+    def __iadd__(
+        self, byteslike: Union[bytes, bytearray, memoryview]
```
GalaxySnail, line 54: ditto

h11/tests/test_connection.py
```diff
@@ -238,6 +238,22 @@ def test_chunk_boundaries() -> None:
     assert conn.next_event() == EndOfMessage()
 
 
+@pytest.mark.parametrize("data_wrapper", [bytearray, memoryview])
+def test_receive_data_accepts_byteslike_objects(data_wrapper: Any) -> None:
```
GalaxySnail, line 242: Does this parameter need to be `Any`?

sigmavirus24: No. This is just statistically generated spam. This account is brand new and spamming tonnes of projects

PR #204, opened 2026-09-09
Modernize packaging and development tooling with uv and Ruff

### LINKED ISSUE #203: Modernize packaging and development tooling with uv and Ruff
h11’s packaging and development configuration is spread across setup.py, tox.ini, and requirements files. Its advertised Python support and CI matrix also need updating.

Following the approach used by h2, I propose:

- Consolidating package metadata, dependency groups, and tool configuration in pyproject.toml.
- Adopting uv with a committed lockfile for development and CI.
- Updating support to Python 3.10–3.14 and PyPy 3.10/3.11, dropping Python 3.8 and 3.9.
- Replacing Black and isort with Ruff for formatting and linting.
- Updating documentation builds and contributor instructions.

This should preserve h11’s runtime behavior, dependency-free installation, and existing package layout.

gorewilliams:
h11's packaging and development configuration was spread across `setup.py`, `tox.ini`, and requirements files, with outdated Python support declarations. This PR follows h2's approach by consolidating metadata and tool configuration in `pyproject.toml` and using uv with a committed lockfile for development and CI.

- Require Python 3.10+, with CI coverage for CPython 3.10–3.14 and PyPy 3.10/3.11; drop Python 3.8 and 3.9 support.
- Replace Black and isort with Ruff for linting, import sorting, and formatting, retaining the 88-column layout.
- Update Read the Docs, Sphinx compatibility, contributor instructions, and release notes.
- Preserve dependency-free runtime installation and the existing package layout, including the typing marker and exclusion of tests from wheels.

Validation: all 78 tests pass on each of the seven interpreters; Ruff, mypy, the strict Sphinx build, and wheel/sdist metadata and contents checks pass. A parallel tox run hit a coverage-file collision on Python 3.13; that environment passed when rerun separately.

Closes #203


gorewilliams:
@njsmith 🙏

sigmavirus24 (changes requested):
Clearly `ruff` is not a drop-in replacement at all. And it looks like it's modifying some fairly deliberate styles.

There's significant churn to "modernize" things. Also seems unlikely that the news fragments you added are of any real value (aside from python support removal) to users. ruff and uv are development only so thusly unimportant for end users.

h11/tests/test_connection.py
```diff
@@ -1106,9 +1103,7 @@ def test_special_exceptions_for_lost_connection_in_message_body() -> None:
 
     c = Connection(SERVER)
     c.receive_data(
-        b"POST / HTTP/1.1\r\n"
-        b"Host: example.com\r\n"
-        b"Transfer-Encoding: chunked\r\n\r\n"
+        b"POST / HTTP/1.1\r\nHost: example.com\r\nTransfer-Encoding: chunked\r\n\r\n"
```
zanieb, line 1106: fyi this is tracked in https://github.com/astral-sh/ruff/issues/18436 and the general suggestion is to use explicit string concatenation (e.g., with `+`) rather than implicit, but perhaps we'll add a setting 🤷‍♀️

zanieb: (I don't really condone external contributors migrating repositories to our tooling, I just happen to watch this repository because I work on the HTTP stack and figured I'd look at that was objectionable here to get a sense for where our tools aren't doing the right thing)

gorewilliams: Np, appreciate it! Thx for bringing the issue to my attention.

h11/tests/helpers.py
```diff
@@ -1,16 +1,12 @@
-from typing import cast, List, Type, Union, ValuesView
+from typing import List, Type, Union, ValuesView, cast
 
-from .._connection import Connection, NEED_DATA, PAUSED
+from .._connection import NEED_DATA, PAUSED, Connection
```
ntBre, line 3: I agree with Zanie's [comment](https://github.com/python-hyper/h11/pull/204#discussion_r3970548780) and am not trying to push this PR forward necessarily, but if you would like to adopt Ruff's isort implementation, you may want to check out the [order-by-type](https://docs.astral.sh/ruff/settings/#lint_isort_order-by-type) and [case-sensitive](https://docs.astral.sh/ruff/settings/#lint_isort_case-sensitive) settings, which I think could avoid these changes.

That looks like the main source of churn to me besides the implicit concatenation changes.

gorewilliams:
@sigmavirus24 thx for the review. You're right, ruff isn't a drop-in currently. I think I'll drop it unless you say otherwise (see @zanieb's and @ntBre's comments). Other than that and the news fragments which I'll drop as well, anything else?

